PR #7, opened 2022-07-12
Sidebar component

belv2c:
## Description

In this PR I created a re-usable and accessible sidebar. It is rendered outside of the root component with `createPortal`.

The sidebar can be closed by: 
- clicking outside of the sidebar
- passing a button as a child and using the onClick attribute to close it
- passing a `title` prop to the sidebar (it will render a header with an 'x' to close it)
- pressing the `escape` key 

The sidebar uses [react-focus-lock](https://github.com/theKashey/react-focus-lock) so that when the sidebar is open, tabbing through focusable elements is contained within the sidebar only (for accessibility).

It also uses custom hooks `useKeyPress` so the sidebar can be closed when the escape key is pressed, and `useOutsideClick` so the sidebar can be closed when a click happens outside of the component.


roblafeve (dismissed):
:shipit:  🚀

roblafeve (approved):
This seems solid 👍

PR #12, opened 2022-07-22
Tabs component

roblafeve:
## Description

Adds a `<Tabs />` component for displaying content.

### Usage

```tsx
<Tabs currentTab="1" onTab={(nextTab) => setCurrentTab(nextTab)}>
  <Tab id="1" label="Tab 1">Content 1</Tab>
  <Tab id="2" label="Tab 2" disabled>Content 2</Tab>
  <Tab id="3" label="Tab 3">Content 3</Tab>
</Tabs>
```

If a user passes a bad initial `id` (one that doesn't exist OR belongs to a disabled tab), we simply throw an error telling them what's wrong.

### 📋 Acceptance Criteria

- [x] Tabs can be composed using both `Tabs` and `Tab` components
- [x] When a "tab" is clicked, its content is displayed
- [x] An "initial tab" index can be set
- [x] An `onTab` function is called with the current index when a tab is selected
- [x] A11y tags have been appropriately applied
- [x] Keyboard events and tabbing is supported
- [x] Supports "disabled" tabs
- [x] Throws errors for bad ID or disabled tab

### 🖼 Demo

https://user-images.githubusercontent.com/157195/180878731-72f1cd64-8c70-4246-8b8c-db80525a6fd5.mov

#### With Disabled Tabs

https://user-images.githubusercontent.com/157195/182234524-a5054b3e-e716-4798-8195-cadff6a9bbbd.mov

### 👩‍🔬 Test Instructions

1. Install dependencies
2. `npm run test:playground`
3. View "Tabs" stories
4. Select various tabs and ensure that corresponding content is displayed5. 
5. Utilize Tab and Arrow Left/Right keys to navigate

## 🔎 Reviewer Checklist

- [x] Merge destination is correct
- [x] Code is correct as understood and conforms to quality standards
- [x] Tests have been added where appropriate (unit, visual, end-to-end)
- [x] Acceptance Criteria have been met

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>


roblafeve:
I'm going to take a fresh stab at this with a focus on simplification.

roblafeve:
Alright, I took a step back and reconsidered a few things. I feel decent about the current state of this. It does now accomodate keyboard events and, as far as I can tell, has the correct accessibility attributes included. Notably, this component is no longer uncontrolled—the user must track state separately.

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Tab } from "./types"
+
+interface TabsProps {
+  /**
+   * We constrain `children` to be element(s) with a label and children
+   * props (e.g. `Tab`). This is only structural, however.
+   **/
+  children: ReactElement<TabProps> | Array<ReactElement<TabProps>>
+  currentTab: number
+  onTab: (index: number) => void
```
thispastwinter: ⛏️ nit pick - I prefer onTabChange here, since this event fires off every time the tab selection changes.

src/components/App/App.tsx
```diff
@@ -41,6 +43,11 @@ export function App() {
         hideLabel={false}
         helperText="this is helper text"
       />
+      <Tabs currentTab={currentTab} onTab={setCurrentTab}>
```
thispastwinter: I think it might make more sense to expose an `id` prop on the Tab component and use the id for handling active states. This would also provide the user with a more meaningful value that would be returned from the onTab callback. I'm just thinking in terms of routing/api requests. Even outside of those considerations, I think reasoning through a value that is returned that they can see assigned to a given tab makes more sense than an index, which feels a little more arbitrary.

src/components/Tabs/Tabs.tsx
thispastwinter, line 14: nit pick - i prefer onTabChange, since this callback is fired every time a tab changes.

thispastwinter: Ah, except that it actually gets called on every tab click, not just when the tab changes. Well maybe that's something to consider, does it need to be called if the tab is already selected? 🤷

roblafeve: @thispastwinter thanks for looking this over! Yeah, I think I tend to agree regarding the naming. I could see `onTabChange` or even just `onChange` making more sense. 👌

Re: Preventing it getting called if already "active"—hmmm, I could see doing this, but I'm trying to weigh the benefits. Do you think it's important?

src/components/App/App.tsx
roblafeve: @thispastwinter yeah, I think that's fair. I had considered just passing the entire tab data back, which would give the user everything. I do like the idea of requiring an `id` prop on `Tab` and using it throughout. I think that's smart. Great suggestion!

src/components/Tabs/Tabs.tsx
thispastwinter: Probably not :)

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export function Tabs({ children, onTab, currentTab }: TabsProps) {
+  // We a list of refs for tablist nodes so we can focus them programmatically
```
albatrocity: @roblafeve _nitpick alert_ 
```suggestion
  // We use a list of refs for tablist nodes so we can focus them programmatically
```

albatrocity (dismissed):

src/components/Tabs/Tabs.tsx
roblafeve: @albatrocity this was very carefully chosen wording and it won't be fixed

belv2c:
@roblafeve I really like that you added all of the roles and attributes to make the tabs fully accessible 👍 . Do you think it would be overkill to add a `disabled` state to the individual tabs? I could see that being a common use case.

roblafeve:
> @roblafeve I really like that you added all of the roles and attributes to make the tabs fully accessible 👍 . Do you think it would be overkill to add a `disabled` state to the individual tabs? I could see that being a common use case.

@belv2c I think that's a good idea! I'll look into it. 👍

src/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,8 @@
+import { ReactNode } from "react"
+
+export interface TabProps {
+  children: ReactNode
+  label: string
```
luiz-chagas, line 6: This property `label` doesn't seem to be used here where it is defined but it's clearly used elsewhere and it breaks the module if I try to remove it.

This to me feels like this type is being "abused" (in a way that is doing more than it is intended to be). `TabProps` reads and feels like it's providing props to the `Tab` component, but it's doing more than that.

I couldn't quite figure out why, but I thought I'd share where I'm going with this.

src/components/Tabs/components/Tab/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Tab } from "./Tab"
+export type { TabProps } from "./Tab"
```
luiz-chagas: Just learned this last week that you can combine these exports into a single statement:
```tsx
export { Tab, type TabProps } from "./Tab"
```

Not that you have to, I just thought it was cool

src/components/Tabs/components/TabPanel/TabPanel.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+interface TabPanelProps {
+  tab: Tab
+}
+
+export function TabPanel({ tab }: TabPanelProps) {
+  const ariaHidden = tab.isSelected === false
+  const className = `${styles.TabPanel}${
+    ariaHidden ? ` ${styles.isHidden}` : ""
+  }`
```
luiz-chagas: On `TabList.tsx` you used `classnames` for a similar idea, here you went with template literals. Any reasons for choosing one over the other or not really? Just curious

src/components/App/App.tsx
```diff
@@ -1,11 +1,13 @@
 import { useState } from "react"
 import { Button } from "../Button"
 import { Modal } from "../Modal"
+import { Tab, Tabs } from "../Tabs"
 import { TextInput } from "../TextInput"
 import "./App.styles.css"
 
 export function App() {
   const [isOpen, setIsOpen] = useState<boolean>(false)
+  const [currentTab, setCurrentTab] = useState(0)
```
luiz-chagas: +1 to using IDs instead of indexes. Indexes work fine, but if you're generating dynamic tabs out of some data ingestion, indexes could end up not being reliable

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export function Tabs({ children, onTabChange, currentTab }: TabsProps) {
+  // We use a list of refs for tablist nodes so we can focus them programmatically
+  const tabListRefs = useRef<(HTMLLIElement | null)[]>([])
+
+  // Sometimes children is an array; sometimes it's a single node—let's make it an array
+  const childrenArray = Array.isArray(children) ? children : [children]
+
+  // Let's make sure the
```
luiz-chagas:
```
// Let's make sure the
```
missing second half of this comment

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+  let currentIndex =
+    currentTab < 0 || currentTab > childrenArray.length - 1 ? 0 : currentTab
```
luiz-chagas: This is _potentially_ dangerous in a way that it could hide a programmer error.

If I try to render this:
```
<Tabs currentTab={999}>
  <Tab label="my tab">...</Tab>
</Tabs>
```

It's clear I'm doing something wrong, would I want for this to break / warn / log and/or do it's best to work?

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  // Now we need to create a list of only enabled children
+  const childrenEnabled = childrenWithIndex.filter((child) => !child.disabled)
+  if (isDisabledCurrent) {
+    /**
+     * If we discovered that the currentTab is disabled, we set the new currentTab to
+     * the index of the first enabled tab we have.
+     **/
+    currentIndex = childrenEnabled[0].index
+  }
```
luiz-chagas: Similar to above, do we need to fix that? Or tell the developer that something is clearly wrong?

luiz-chagas (commented):
This is looking good! I haven't had the chance to go through everything, but since I'm about to break here for lunch I thought I'd leave a partial review for now.

src/components/Tabs/components/Tab/Tab.tsx
roblafeve: Yep, totally understand. Since we are dynamically reading children passed to `<Tabs />`, `<Tab />` is a convenient way to pass valid children in. We definitely _need_ label, but it's read somewhere else (in `Tabs`). Feels a little odd, but given the current API, it's necessary.

src/components/Tabs/components/Tab/index.ts
roblafeve: Oh, that's cool. I didn't know you could do that—I like it!

src/components/Tabs/components/TabPanel/TabPanel.tsx
roblafeve: Great question. @albatrocity had suggested that we might draw the line on dependencies and exclude the usage of libs like `classnames` and instead manually do this sort of thing—so since this case was pretty simple I went for it manually, BUT haven't gone back and reworked the spot where I'm still using `classnames`. What are your thoughts—is it worth trying to avoid using it?

src/components/App/App.tsx
roblafeve: I haven't converted this over to IDs just yet, but that's the next step 👍

src/components/Tabs/Tabs.tsx
roblafeve: Yeah, I wasn't sure how to best handle this. It's definitely simpler to just tell the user to fix the problem. It would definitely simplify the implementation considerably.

roblafeve:
Oof—merge rebase was a mistake in this case. 😞

src/components/Tabs/Tabs.tsx
roblafeve: I've switched to just throwing errors when the `id` doesn't exist or it belongs to a disabled tab.

src/components/Tabs/components/TabPanel/TabPanel.tsx
roblafeve: I've removed the dependency on `classnames` from `TabList.tsx` as well.

README.md
```diff
@@ -25,7 +25,7 @@
 - ~~Loader~~
 - ~~Toast~~
 - Banner
-- Tabs
+- Tabs (in progress)
```
thispastwinter: Might wanna just cross it out since we're nearing the finish line 😄

src/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onKeyDown,
+  tab,
+  refs,
+}: TabListItemProps) {
+  const className =
+    styles.item +
+    (tab.selected ? " " + styles.selected : "") +
+    (tab.disabled ? " " + styles.disabled : "")
+
+  const handleClick = () => onClick(tab)
```
thispastwinter: ⛏️ nitpick alert - Is it necessary to pass in the whole tab object here? It looks like the only time you extract a property from this object is when you pass the `id` from the callback to your `onTabChange`. Seems like you could just pass the id here, unless you had a vision for future implementation.

src/components/Tabs/components/TabList/TabList.tsx
```diff
@@ -0,0 +1,69 @@
+import { KeyboardEvent, MutableRefObject } from "react"
```
thispastwinter, line 1: Might be worth adding stories for your TabList and TabListItem.

src/components/Tabs/components/TabList/TabList.tsx
roblafeve: Nah, I think I was just hedging my bets. Passing the `id` makes most sense now. 👍

src/components/Tabs/components/TabList/TabList.tsx
roblafeve: Hmmm, I started down that path, but it felt like a bit much. Do you think it would be useful to, say, style those pieces in the abstract?

README.md
roblafeve: Doh! Yeah, I suppose I should have stared with it crossed off. 🤣

src/components/Tabs/components/TabList/TabList.tsx
thispastwinter: I think stories or somehow breaking down the api a bit more in the docs. Maybe just some urls that point to respective components for changing the look of the tab items.

roblafeve: Yeah, I think fleshing out the nested READMEs and linking to them from the main one would be helpful. Good callout!

src/components/Tabs/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Directory Structure
+
+```
+Tabs
+├── components 👈 sub-components
+│   ├── Tab
+│   ├── TabList
+│   └── TabPanel
+├── README.md 👈 you are here
```
luiz-chagas: 😄

src/components/Tabs/components/TabList/TabList.tsx
roblafeve: @thispastwinter take a look

luiz-chagas (changes requested):
When a tab is `disabled`, the `TabListItem` does not disable interactions, which I believe not to be the expected behavior.

roblafeve:
> When a tab is `disabled`, the `TabListItem` does not disable interactions, which I believe not to be the expected behavior.

@luiz-chagas this was a test. You passed! Just kidding. I refactored some styling and had a typo—of course. It's fixed now and I added a test to cover this.

src/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { Tab as TabType } from "../../types"
+
+export interface TabProps {
+  children: ReactNode
+  label: string
+  id: TabType["id"]
+  disabled?: boolean
+}
+
+export const Tab = ({ children }: TabProps) => <>{children}</>
```
luiz-chagas, line 11: This is the component that put me down this rabbit hole.

You can re-write this component as `export const Tab: FC<TabProps> = () => null` and the tabs still work the exact same way!

luiz-chagas (approved):
This was a LOT of work, good stuff. Using Tabs `children` to generate data to then generate other components is pretty dope 🚀 

One line set me off on a rabbit hole and it lead to interesting findings:

I believe the `Tab` and `TabPanel` components are essentially the same component, `TabPanel` seems to be the internal version of `Tab` with an extra `selected` property. I tried to write a version of this that combines both ideas in a single thing and it's very interesting if anything. If you don't mind `selected` leaking to the main `Tab` component, it's pretty much a 1:1 replacement, if you want to keep `selected` internalized (a good thing IMO), then it requires a little bit more "magic" to add that property on the fly. (I ended up using Context). I'll open a PR with my approach just for reference in case you're curious.

belv2c (approved):
Nice work on this! 🎉

thispastwinter:
Well done!

thispastwinter (approved):

src/components/Tabs/components/Tab/Tab.tsx
roblafeve: >I believe the Tab and TabPanel components are essentially the same component

Ha ha, you're completely right! I'll need some time to look through your solution, but if both require some magic to prevent `selected` from leaking, the question then is, "which requires less magic?"

There's definitely a part of me that wants to go back and refactor this whole thing now that I have a better sense of what's required—taking into account what we've learned. BUT, I think I'll merge this for now and open a new PR that addresses those changes specifically to make reviewing all that a bit more manageable.

PR #16, opened 2022-07-25
Feature/toast

### LINKED ISSUE #29: Toast

thispastwinter:
## Description

Adds both a Toast component and Toast system (Toaster) for handling the displaying of toasts.

### 🔗 Reference

- [LINK_NAME](LINK_URL)

### 📋 Acceptance Criteria

- [ ] CRITERION_HERE

### 🖼 Demo

https://user-images.githubusercontent.com/46500734/181075942-4a78db4c-3d73-429b-a523-c110c30ee8b8.mov

### 👩‍🔬 Test Instructions

1. Start storybook via `npm run test:playground`
2. Comb through each story to see a representation of each Toast status type
3. Play with the system in the Components/Toast/ProviderExample.

## 🔎 Reviewer Checklist

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>


src/components/Toast/ToasterContext.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const addToast: ContextProps["addToast"] = useCallback(
+    (toast) => {
+      const newToast: ToastType = { ...toast, id: v4() }
+      if (behavior === "stack") {
+        if (position.vertical === "center" || position.vertical === "top") {
+          setToasts((toasts) => [newToast, ...toasts])
+        } else {
+          setToasts((toasts) => [...toasts, newToast])
+        }
```
roblafeve: 💭 I wonder if it would be simpler to control the stack order using `flex-direction: column-reverse` as I assume you're taking these properties into account for other styling.

src/components/Toast/ToasterContext.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const removeToast = useCallback((id: string) => {
+    setToasts((toasts) => toasts.filter((toast) => toast.id !== id))
+  }, [])
+
+  const removeOldest = useCallback(() => {
+    if (toasts.length === 0) {
+      return
+    }
+    setToasts((toasts) => toasts.slice(1))
+  }, [toasts])
```
roblafeve, line 69: Related to my comment on line 66; it seems that the oldest toast might not be at index 0

src/components/Toast/ToastPlayground.tsx
```diff
@@ -0,0 +1,64 @@
+import { Button } from "../Button"
+import { useToaster } from "./useToaster"
+
+export function ToastPlayground() {
+  const { addToast, removeAll } = useToaster()
```
roblafeve: ❤️  I love the use of the term "toaster"

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function Toast({ action, message, status = "success" }: ToastProps) {
+  return (
+    <div
+      data-testid={`toast_${message}`}
+      role="alert"
+      aria-live={status === "error" ? "assertive" : "polite"}
+    >
+      <div
```
roblafeve: I am likely ignorant here—but could you combine the outer and inner divs' attributes and remove the outer div altogether?

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,32 @@
...
+        aria-label={`a ${status} toast`}
```
roblafeve: Come to think of it, I'm not sure we need three separate divs to support the intended structure. 🤷

src/components/Toast/README.md
```diff
@@ -0,0 +1,21 @@
+# `<Toast />`
+
+All Toast related dependencies are contained within this folder, with the exception of 3rd party libraries and a shared hook `useKeyPress`.
+
```
roblafeve, line 4: 💭 Some recent work has included a ["Props" table](https://github.com/cremalab/crema-components/blob/9d0ff9dfa48550c085e474ca211491598a9edc24/src/components/Tabs/README.md#api)—what are your thoughts? Is that too much?

src/components/Toast/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Directory Structure
```
roblafeve, line 5: 👍 I like the thoroughness here. It reminds me—I didn't include sub-components in mine. 😬

src/components/Toast/utils/getToastTransitionStyles/getToastTransitionStyles.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * getToastTransitionStyles
+ * ------------------------------------------------
+ * This util is used to generate a transition styles object
+ * based on animation duration, transition state, position,
+ * and behavior.
+ */
```
roblafeve: I like the inclusion of these comments. Hot tip—if you place the comment directly above the function definition itself, then it will be displayed when hovering over it at the call-site.

src/components/Toast/utils/getToastPositionStyles/getToastPositionStyles.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * getToastPositionStyles
+ * ------------------------------------------------
+ * This util is used to generate a styles object that
+ * dictates the toast position based on the Position
+ * arguments.
+ */
```
roblafeve, line 14: See previous comment.

src/components/Toast/utils/ToasterConfig/ToasterConfig.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Position } from "../../types"
+
+export interface Config {
+  duration: number
+  animationDuration: number
+  position: Position
+  behavior: "stack" | "replace"
+}
+
+export class ToasterConfig {
```
roblafeve: I'd love to hear your reason for going with a class vs a factory function. Just curious—this looks like it does what's intended. 👌

roblafeve (commented):
This is really shaping up 🎉

src/components/Toast/Toast.tsx
thispastwinter: You're right. I think this was a byproduct of refactoring.

src/components/Toast/utils/ToasterConfig/ToasterConfig.ts
thispastwinter: It was mostly an excuse to use a class, but the idea here was that it omits the possibility of passing an inline configuration object to the provider and ensures that a config object is always passed, which makes the developer more likely to know that it's an option to override those defaults...outside of documentation.

thispastwinter: If there is a better way without the use of a class, I'm totally open to it. I also realize it's not that big of a deal to pass an object inline, and for that matter you could set the object to a variable and pass in that variable, I just liked the patter that it forced.

src/components/Toast/ToasterContext.tsx
thispastwinter: Yeah, you're probably right, since maintaining ordering is really only a visual perk, and isn't necessary from a functionality standpoint. If we programmatically remove a toast via it's handleclose callback, it doesn't matter much if we maintain the proper order, since it's only ever removing a singular toast by its given id.

src/components/Toast/README.md
thispastwinter: I feel like with how few props it has, it might be, but I'm happy to follow the trend.

roblafeve: Something we should probably standardize and add to the hygen templates if we want to go down that path.

src/components/Toast/ToasterContext.tsx
thispastwinter: So actually, I think that we need to maintain `flex-direction: column` to ensure that items appear to enter from the appropriate direction, otherwise they don't look like they're "stacking". I think the fix might be something like this however.

```
const isTopOrCenter =
    position.vertical === "center" || position.vertical === "top"
    
const removeOldest = useCallback(() => {
    if (toasts.length === 0) {
      return
    }
    if (isTopOrCenter) {
      setToasts((toasts) => toasts.slice(0, toasts.length - 1))
    } else {
      setToasts((toasts) => toasts.slice(1))
    }
  }, [toasts, isTopOrCenter])
  
const addToast: ContextProps["addToast"] = useCallback(
  (toast) => {
    const newToast: ToastType = { ...toast, id: v4() }
      if (behavior === "stack") {
        if (isTopOrCenter) {
          setToasts((toasts) => [newToast, ...toasts])
        } else {
          setToasts((toasts) => [...toasts, newToast])
        }
      } else {
        setToasts([newToast])
      }
    },
    [behavior, isTopOrCenter],
  )
```

thispastwinter: But maybe this is smelly and there is still a more elegant solution.

src/components/Toast/utils/ToasterConfig/ToasterConfig.ts
roblafeve: Yep, that makes sense. I think at least having something produce a default config is a good idea—but being able to pass a simple object sounds appealing as well. So maybe just a function that produces this config would be enough?

src/components/Toast/ToasterContext.tsx
roblafeve: @thispastwinter yeah, this feels pretty complex. So `flex-direction: column` and `flex-direction: column-reverse` don't work identically aside from the order? I haven't played with it, but that's surprising to me. 🤔

thispastwinter: Let me think through this further and see what I can do 😄

thispastwinter: No, disregard. I had my logic mixed up earlier when I attempted to just use flex.

src/components/Toast/utils/ToasterConfig/ToasterConfig.ts
thispastwinter: I think if we go that route, I'd almost prefer just eliminating a util altogether and having the prop just define the object shape. It kinda goes against the pattern I was trying to create, but if we are going for simplicity/less code we can totally just delete all that.

package.json
```diff
@@ -3,11 +3,15 @@
...
   "dependencies": {
+    "@types/react-transition-group": "^4.4.5",
+    "@types/uuid": "^8.3.4",
     "classnames": "^2.3.1",
     "react": "18.1.0",
     "react-dom": "18.1.0",
     "react-focus-lock": "^2.9.1",
     "react-scripts": "5.0.1",
+    "react-transition-group": "^4.4.2",
+    "uuid": "^8.3.2",
```
luiz-chagas: Seems excessive to bring in `uuid` just to generate an ID for toasts. 

Could toast ids be a number? Maybe starting at `1` and going up 1 by 1?

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,29 @@
...
+      aria-live={status === "error" ? "assertive" : "polite"}
+      aria-label={`a ${status} toast`}
+      data-type={status}
+      className={styles.toast}
+    >
+      <div className={styles.toastContent}>
+        <div>{message}</div>
+        {action ? (
+          <div className={styles.toastActionContainer}>{action}</div>
+        ) : null}
```
luiz-chagas: Not a request, just a question:

I see this pattern quite often, is this any different than:
```tsx
{action && <div className={styles.toastActionContainer}>{action}</div>}
```

Ternary operators seem harder to read for me (probably because I have a bad attitude towards them to begin with) but not sure if these statements are equivalent

src/components/Toast/ToasterContext.tsx
```diff
@@ -0,0 +1,136 @@
...
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react"
+import { createPortal } from "react-dom"
+import { v4 } from "uuid"
+import { Transition, TransitionGroup } from "react-transition-group"
+import { useKeyPress } from "../../hooks/useKeyPress"
+import { Config, Toast as ToastType } from "./types"
```
luiz-chagas: These types are only used in this file, could they be defined here instead? 

(I can't find a good reason to keep `types.ts` type of file around, types are everywhere anyway, no need to make an extra file for _some_ types)

src/components/Toast/ToasterContext.tsx
```diff
@@ -0,0 +1,136 @@
...
+    position = defaultConfig.position,
+    animationDuration = defaultConfig.animationDuration,
+    duration = defaultConfig.duration,
+    behavior = defaultConfig.behavior,
+  } = config
+
+  const [toasts, setToasts] = useState<ToastType[]>([])
+  const timer = useRef<NodeJS.Timeout>()
+  const isTopOrCenter =
+    position?.vertical === "top" || position?.vertical === "center"
```
luiz-chagas: It looks like position is never `undefined`, no need for `?`

src/components/Toast/utils/getToastTransitionStyles/getToastTransitionStyles.ts
```diff
@@ -0,0 +1,42 @@
...
+export const getToastTransitionStyles = ({
+  animationDuration,
+  transitionStatus,
+}: Args) => {
+  const defaultStyle: CSSProperties = {
+    transition: `transform ${animationDuration}ms ease-in-out`,
+    width: 300,
```
luiz-chagas: This seems to be the line of code that changes the `width` of a toast, I would think adjusting width/height would be "base cases" but it's placed all the way down here.

If you try to change the width of  the main `.toast` css class (an obvious place for this kind of styling), this 300 causes the toast to be rendered off-center.

This isn't wrong but feels complicated to do something simple.

src/components/Toast/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  horizontal: "left" | "center" | "right"
+}
+
+export type Status = "success" | "info" | "error" | "warning"
+
+export type Toast = {
+  id: string
+  message: string
+  status: Status
+  action?: (handleClose: () => void) => ReactNode
```
luiz-chagas: This API (`Toast['action']`) feels maybe a bit rough.

This is an argument that gets passed as a param to a function (addToast), that expects a callback function that will then return some JSX and maybe inject the `handleClose` argument.

I see why you needed this but also wonder if you could have made your life easier by taking a simplistic approach and not taking in custom JSX.

One idea:
```ts
action: {
  type: 'dismiss' | 'dismissAll'
  text: string
}
```

(Or replace action with `dismissText` or `dismissAllText`)

It would let you go from this function call:
```tsx
addToast({
  message: 'Success Toast',
  status: 'success',
  action: (handleClose) => (
    <Button ariaLabel="close toast" onClick={handleClose}>
      Dismiss
    </Button>
  ),
});
```

To this:
```tsx
addToast({
  message: 'Success Toast',
  status: 'success',
  action: {
    text: 'Dismiss',
    type: 'dismiss',
  },
});
```

You lose changeability (all your action buttons would look the same, which could be good for cohesion anyway) but gain a little bit in ergonomics. ⚖️

luiz-chagas (commented):
This is looking good! Making a toast system isn't trivial and you're pretty much there.

I have 2 good questions (about width and action) and a bunch of smaller things that you can even ignore lol.

src/components/Toast/types.ts
thispastwinter: I like this. I agree that this simplifies things.

src/components/Toast/utils/getToastTransitionStyles/getToastTransitionStyles.ts
thispastwinter: Yeah, that makes a lot of sense. I think I was just moving through so many iterations and refactors with all of this, that some weird artifacts were born from that. I'll just move that over to the actual Toast stylesheet and it will be more clear where to change those properties.

src/components/Toast/ToasterContext.tsx
thispastwinter: Good catch, that was leftover from a previous implementation where those could have been undefined.

src/components/Toast/Toast.tsx
thispastwinter: I think I've just gotten into this habit because I have had issues where truthy values make it through and render as text on the screen, but maybe that speaks more of my mishandling of truthy values.

package.json
thispastwinter: Yeah, that's a good callout, especially given the lifespan of a toast.

roblafeve (dismissed):
I think this is looking good 👍

thispastwinter:
> I think this is looking good 👍

Thanks! I did notice this morning I also need to remove the uuid type defs. I'd still like to get @luiz-chagas 👁️ on my changes as many of my recent changes were aimed at resolving some of his feedback.

luiz-chagas (approved):
Looks good to me 💪🏼 🚀

PR #21, opened 2022-08-02
Add a Chip component

### LINKED ISSUE #24: Chip
Addressed by #21

blaketarter:
## Description

This PR adds a basic chip component used to show "chip" or "tag" like information. The component supports `onSelect` and `onClose` actions with keyboard support and a replaceable `closeElement` prop.

### 🖼 Demo

Screen Shot 2022-08-02 at 4 04 23 PM

https://user-images.githubusercontent.com/6101761/182473488-550c586e-90f7-4eab-8c55-5ea24dae883f.mov

## 🔎 Reviewer Checklist

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>

### DIFF
--- a/src/components/Chip/Chip.module.css
+++ b/src/components/Chip/Chip.module.css
@@ -1,0 +1,3 @@
+.Chip {
+  display: flex;
+}
--- a/src/components/Chip/Chip.module.css
+++ b/src/components/Chip/Chip.module.css
@@ -1,3 +1,4 @@
-.Chip {
+.container {
   display: flex;
+  cursor: pointer;
 }
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -1,0 +1,50 @@
+import { ReactNode } from "react"
+import styles from "./Chip.module.css"
+
+export interface ChipProps {
+  children?: ReactNode
+  onSelect?: () => unknown
+  onClose?: () => unknown
+  closeElement?: ReactNode
+}
+
+export function Chip({
+  children,
+  closeElement = <span>&times;</span>,
+  onSelect,
+  onClose,
+}: ChipProps) {
+  return (
+    <div className={styles.Chip}>
+      <div
+        onClick={onSelect}
+        // these next 3 props are important for a11y
+        role={onSelect ? "button" : undefined}
+        tabIndex={onSelect ? 0 : undefined}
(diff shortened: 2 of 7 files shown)


blaketarter:
I'm torn on how much styling to include with something like this. I could add some more default styles to make it look better, or I could even remove all styles entirely. Anyone have any thoughts?

albatrocity (dismissed):
Looks good to me. I'd say having an css module entry point like you have is enough to pave the way for "real" styling.

src/components/Chip/Chip.module.css
```diff
@@ -0,0 +1,3 @@
+.Chip {
```
roblafeve: Something I've noticed is that if we use the component name as the wrapper class, then css modules doubles up on it. I wonder if we should use a convention like `container` or something. That's what I did. 🤷 

Screen Shot 2022-08-02 at 5 21 11 PM

src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,50 @@
+import { ReactNode } from "react"
+import styles from "./Chip.module.css"
+
+export interface ChipProps {
+  children?: ReactNode
```
roblafeve: Is `children` being optional a realistic scenario?

roblafeve (dismissed):
Nice work!

belv2c (commented):
Nice work! Since the chip is selectable and the closeElement is clickable, do you think it makes sense to add `cursor: pointer` to the css for these?

thispastwinter:
> Nice work! Since the chip is selectable and the closeElement is clickable, do you think it makes sense to add `cursor: pointer` to the css for these?

+1

thispastwinter (commented):
~~I personally think styling things a bit wouldn't hurt, if only to ensure that each respective element has a class tied to it. I'm trying to think of what would make this the most valuable if I were to pull it into a project. Knowing I could use things as is and quickly make some changes to various classes to get it to look good would be a huge boon.~~

I take that back, because you've basically given a great deal of freedom to enable someone to style it however they want by having a `closeElement` prop and a container that accepts a `ReactNode` child.

roblafeve (approved):
👌 :shipit:

belv2c (approved):
🚢

roblafeve:
resolves #24

PR #22, opened 2022-08-02
Combine Tab and TabPanel into a single thing

### LINKED ISSUE #25: Tabs

luiz-chagas:


luiz-chagas:
CI is complaining that `Tab.tsx` doesn't exist but that's how it's supposed to be, tests pass locally, not sure what's going on but also not worth investigating here

PR #37, opened 2022-08-03
Breadcrumbs component

### LINKED ISSUE #23: Breadcrumbs

belv2c:
# Breadcrumb

Displays a list of links to show users where they are within an application.

### Keyboard Interaction

Not applicable.

### WAI-ARIA Roles, States, and Properties

- The links are contained in an ordered list within a `<nav>` element
- The `<nav>` element has the `aria-label` attribute set to `"breadcrumbs"`
- The last link in the list represents the current page, and must have `aria-current` set to `"page"`

### 🔗 Reference

- [ARIA Authoring Practices Guide breadcrumb navigation example](https://www.w3.org/WAI/ARIA/apg/example-index/breadcrumb/index.html)


thispastwinter:
🥖 🍞

src/components/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    ]
+    const children = <BreadcrumbItem links={breadcrumbLinks} />
+
+    render(<Breadcrumb>{children}</Breadcrumb>)
+
+    expect(screen.getByText("Home")).toBeInTheDocument()
+    expect(screen.getByText("About")).toBeInTheDocument()
+
+    const lastLink = screen.getByText("About")
+    expect(lastLink).toHaveAttribute("aria-current", "page")
```
albatrocity, line 27: I suggest breaking these out into separate tests. I try to keep 1 expect per test as long as stand-up overhead isn't huge. In the current scenario, if someone removed the `aria-current` attribute the `renders children` test would fail, which isn't very helpful at a glance.

src/components/Breadcrumb/README.md
```diff
@@ -0,0 +1,12 @@
+# `<Breadcrumb />`
+
+DESCRIPTION_HERE
```
albatrocity: Might as well outline the available props here so that it doesn't look like an incomplete component.

albatrocity (changes requested):
A couple nit-picks. I also wanted to ask if it's outside of the scope of this component to have an `onClick` handler for breadcrumb items, since the steps may not always correspond to a route. Along the same lines it might be good to have a render prop available in case the element needs to be something other than an `a` (thinking of `@reach/router` links or buttons).

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+  href: string
+  isCurrent: boolean
+  label: string
+}
+
+interface Props {
+  links: LinkProps[]
+}
+
+export function BreadcrumbItem({ links }: Props) {
```
luiz-chagas: @belv2c Looking good overall! 

Maybe I'm getting this wrong but the story for this component suggests that the ideal utilization is to have:
```
1 Breadcrumb -> 1 BreadcrumbItem -> Many links
```

I would think that a breadcrumb would have a collection of links (or items), somewhat similar to how `Tabs` have a collection of `Tab` components.

Is there a case for having a `BreadcrumbItem` take a collection of `link`s? For example, would it make sense to have several `BreadcrumbItem`s and each one of them have their own list of `links`? 🤔 

I'm thinking the straightforward path is to have the `BreadcrumbItem` _be_ one link, so then you can enforce that a `Breadcrumb` _should_ only have `BreadcrumbItem`s as children. (as opposed to any `ReactNode`)

Sorry if this was very confusing, I can try to explain my point better

roblafeve: I agree with what @luiz-chagas is pointing out here. I wouldn't expect an "item" to produce a "list"—I expect a "list" to produce "items". If that makes sense... 😆

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,34 @@
+import styles from "./Breadcrumb.module.css"
+
+interface LinkProps {
+  href: string
+  isCurrent: boolean
```
roblafeve, line 5: I'm wondering if `isCurrent` should be a computed property since, as I understand it, the last item in the list is the current page. Is there a scenario that I'm not considering?

src/components/Breadcrumb/BreadcrumbItem.tsx
belv2c: @luiz-chagas @roblafeve This makes sense - I was struggling with some naming variations `links` `link` `item` `items` 😂. I've set aside some time tomorrow to resolve this PR feedback!

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,14 @@
+import { ReactNode } from "react"
+import styles from "./Breadcrumb.module.css"
+
+interface Props {
+  children: ReactNode
```
roblafeve: I'm wondering if it makes sense for this component to accept children since we are configuring links with an aray (on "BreadcrumbItem"). If we stick with that form of configuration, I would suggest passing it here instead of to a child component (I don't think we'd need the child component—exposed at least).

That said, if we want to switch to the composable style that we've used for Tabs, then we'll at least want to constrain children to only accept `BreadcrumbItem`s.

roblafeve (commented):
Loving where this is going! However, it does seem like we might want to reconsider the current structure a bit. What do you think?

src/components/Breadcrumb/Breadcrumb.test.tsx
belv2c: @albatrocity thanks for this feedback! I think it would be nice to provide the option to render an `a` or `button` element here. I did consider adding some built in router configuration but didn't want to overcomplicate the component. For example, what if someone consuming the Breadcrumb is using Next.js and needs to render `Link` from `next/link` instead of `@reach/router`?

albatrocity: @belv2c yes sorry that I wasn't clear: I'm suggesting being able to pass in a component to render as a Breadcrumb Item as opposed to passing in a `href` prop. That way you could pass in a `Link` from `next/link`. I agree that built-in router config seems out of scope.

I was thinking something like this:

```tsx
import { ReactNode } from "react"
import styles from "./Breadcrumb.module.css"

interface LinkProps {
  href: string
  isCurrent: boolean
  label: string
}

interface Props {
  links: LinkProps[]
  renderItem?: ({
    label,
    index,
    isCurrent,
  }: {
    label: string
    index: number
    isCurrent: boolean
  }) => ReactNode
}

export function BreadcrumbItem({ links, renderItem }: Props) {
  return (
    <>
      {links.map((link, index) =>
        renderItem ? (
          renderItem({ ...link, index })
        ) : (
          <li className={styles.BreadcrumbListItem} key={index}>
            <a
              href={link.href}
              className={styles.link}
              aria-current={link.isCurrent ? "page" : "false"}
            >
              {link.label}
            </a>
            {!link.isCurrent && (
              <span aria-hidden="true" className={styles.Separator}>
                /
              </span>
            )}
          </li>
        ),
      )}
    </>
  )
}

```

That way you could do something like this for custom router support:

```tsx
    const breadcrumbLinks = [
      {
        href: "#",
        label: "Home",
        isCurrent: false,
      },
      {
        href: "#",
        label: "About",
        isCurrent: true,
      },
    ]

return <Breadcrumbs renderItem={({ href, label, isCurrent}) => <Link to={href}>{label}</Link>} />
```

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,22 @@
+import styles from "./Breadcrumb.module.css"
+import { BreadcrumbItem, LinkProps } from "./BreadcrumbItem"
+
+interface Props {
+  links: LinkProps[]
```
roblafeve: Instead of reusing `LinkProps` here and exposing `isCurrent`—an implementation detail—to the caller, I would suggest creating a separate `Link` interface that purely represents what the caller needs to know about a link (omitting `isCurrent`).

Then, in your `LinkProps` (or `BreadcrumItemProps`), you could make `isCurrent` required. This way we eliminate confusion at the caller level, and reduce the ability for misusing `BreadcrumbItem` when someone is adjusting the implementation internally.

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,24 @@
+import styles from "./Breadcrumb.module.css"
+
+export interface LinkProps {
+  href: string
+  label: string
+  isCurrent?: boolean
+}
```
roblafeve: As I mentioned previously, this is what I envision happening here.
```suggestion
export interface BreadcrumbItemProps {
  link: Link
  isCurrent: boolean
}
```

src/components/Breadcrumb/Breadcrumb.module.css
```diff
@@ -0,0 +1,22 @@
+.BreadcrumbContainer {
```
roblafeve: ⛏️ Because css modules prefix the class with the module name, we've begun to adopt simpler, context-specific class names. For instance, this one would become:
```suggestion
.container {
```

src/components/Breadcrumb/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Breadcrumb } from "./Breadcrumb"
+export { BreadcrumbItem } from "./BreadcrumbItem"
```
roblafeve: Should we be exporting this still?
```suggestion
```

roblafeve (commented):
Great work! I think this is feeling better from the perspective of the caller. I just have a few thoughts and ideas that might clean things up a tad.

belv2c:
@roblafeve @albatrocity I pushed a commit with some changes for feedback mentioned above. Let me know what you think whenever you have time to take a look!

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,19 +1,38 @@
...
+
+export interface LinkElement {
+  renderItem?: ({
+    item,
+    isCurrent,
+  }: {
+    item: Link
+    isCurrent: boolean
+  }) => ReactNode
+}
```
roblafeve: ✋ I think a caller might be thrown off by the nesting of the render prop within an object. Any reason we wouldn't do this:
```suggestion
export type RenderItem = ({
    item,
    isCurrent,
  }: {
    item: Link
    isCurrent: boolean
  }) => ReactNode
```

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,19 +1,38 @@
...
 
 interface Props {
-  links: LinkProps[]
+  links: Link[]
+  linkElement?: LinkElement
```
roblafeve: See previous:
```suggestion
  renderItem?: RenderItem
```

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -1,24 +1,29 @@
...
   return (
-    <li className={styles.BreadcrumbItem}>
-      <a href={href} className={styles.link} aria-current={ariaCurrent}>
-        {label}
-      </a>
-      {!isCurrent && (
-        <span aria-hidden="true" className={styles.Separator}>
-          /
-        </span>
+    <li className={styles.list}>
```
roblafeve: ⛏️ It feels odd to label a list item as a "list"...
```suggestion
    <li className={styles.listItem}>
```

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -1,24 +1,29 @@
...
+      {linkElement?.renderItem ? (
+        linkElement.renderItem({ item: link, isCurrent })
+      ) : (
+        <a href={link.href} aria-current={ariaCurrent}>
```
roblafeve: 💭 I wonder if it would be at all useful to pass down the class for use on a passed item? Maybe a bad idea. 🤷 
```suggestion
      {linkElement?.renderItem ? (
        linkElement.renderItem({ item: link, isCurrent, className: styles.link })
      ) : (
        <a href={link.href} aria-current={ariaCurrent} className={styles.link}>
```

roblafeve (changes requested):
Only one change request here (✋), everything else is just me being way too granular.😳 Please don't hesitate to remove me from the code reviewers list if I'm being too much.😆 I think we're on the final yard-line! 🏈

src/components/Breadcrumb/Breadcrumb.tsx
belv2c: @roblafeve I like this!

src/components/Breadcrumb/BreadcrumbItem.tsx
belv2c: @roblafeve I think this should help simply the css too

roblafeve: Ah, yeah—explicitly styling by class vs nested element... yeah, I think so

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -8,31 +8,31 @@ export interface Link {
...
-  }) => ReactNode
-}
+export type RenderItem = ({
+  item,
+  isCurrent,
+  className,
+}: {
+  item: Link
+  isCurrent: boolean
+  className?: string
```
roblafeve: ⛏️ It looks like we always pass this in?
```suggestion
  className: string
```

src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -1,25 +1,29 @@
...
+    <li className={styles.listItem}>
+      {renderItem ? (
+        renderItem({ item: link, isCurrent, className: styles.listItem })
       ) : (
-        <a href={link.href} aria-current={ariaCurrent}>
+        <a
+          href={link.href}
+          aria-current={ariaCurrent}
+          className={styles.listItem}
+        >
```
roblafeve: ⛏️ It seems odd that `styles.listItem` is applied to the `li` and the `a` (or rendered item). 🤔

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -8,31 +8,31 @@ export interface Link {
...
-  linkElement?: LinkElement
+  renderItem?: RenderItem
 }
 
-export const Breadcrumb = ({ links, linkElement }: Props) => {
+export const Breadcrumb = ({ links, renderItem }: Props) => {
   return (
     <nav aria-label="breadcrumbs" className={styles.container}>
-      <ol className={styles.list}>
+      <ol>
```
roblafeve: ⛏️ I get that the container styles "cascade", but I wonder if indirectly styling the `ol` via its parent could make it harder to reason about in the future.

roblafeve (dismissed):
🚀 Great work, and thanks for rolling with the feedback!

I noticed a few minor things, but totally your call on whether they matter or not 🤷

src/components/Breadcrumb/Breadcrumb.tsx
belv2c: @roblafeve yeah, i agree. good catch!

src/components/Breadcrumb/Breadcrumb.tsx
belv2c: @roblafeve I think this was a muscle memory thing 😂

luiz-chagas (commented):
I have missed a lot of context from recent reviews/revisions but this is looking good!

What were the reasons that lead you to pursue this particular design:
```tsx
const breadcrumbLinks = [
  {
    href: "#",
    label: "About",
  },
]

render(
      <Breadcrumb
        links={breadcrumbLinks}
        renderItem={(props) => <Link href={props.item.href}>{props.item.label}</Link>}
      />,
    )
```

As opposed to exposing `BreadcumbItem`? Possible Example:
```tsx
render(
      <Breadcrumb>
        <BreadcrumbItem><Link href="#">Home</Link></BreadcrumbItem>
        <BreadcrumbItem><Link href="#">About</Link></BreadcrumbItem>
      </Breadcrumb>
    )
```

Either approach seems to do the job but I'm curious to learn why one over the other (if any)

roblafeve:
> What were the reasons that lead you to pursue this particular design

For some reference, there was some discussion here: https://github.com/cremalab/crema-components/issues/23

belv2c:
@luiz-chagas my goal was to provide _some_ flexibility via the render prop (thanks @albatrocity) while also enforcing component structure and attributes for accessibility.

I think that the component in its' current state is a good compromise between not having to write as much code for the consumer (thanks @roblafeve) while also allowing for customization of the link element via `renderItem`.

roblafeve (approved):
:shipit: 🚀

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { BreadcrumbItem } from "./BreadcrumbItem"
+
+export interface Link {
+  href: string
+  label: string
+  onClick?: () => void
+}
+
+export type RenderItem = ({
+  item,
```
thispastwinter, line 12: A bit of a nit-pick ⛏️. Would `link` maybe be a better name here? It feels weird to me that you have a `renderItem` callback that returns an object with an `item` property.

albatrocity (approved):

thispastwinter (approved):
Feel free to completely disregard my ramblings. Solid work! 🥳

src/components/Breadcrumb/Breadcrumb.tsx
thispastwinter: That said, I guess we are rendering an `item` and said item is given to us here for us to then render how we wish. Maybe just ignore me altogether. Ha.

belv2c: @thispastwinter I was trying to keep the name of this fairly generic since the item rendered can really be anything - a button, router link, anchor tag, etc.

thispastwinter: Yeah, totally. It makes more sense to me now and feels less weird now. I think I just saw how the item's type was assigned to your `Link` type def and it got me thinking about whether it should be called item or not.

Issue #28, opened 2022-08-03
Tooltip


roblafeve:

roblafeve:
@thispastwinter it could be cool to provide your initial plan here. It's something I wish I had done with the Tabs component. Just an idea, though. 🤷

albatrocity:
Unless we're trying to avoid external dependencies, [Popper](https://popper.js.org/react-popper/v2/) can be helpful for dynamic positioning.

PR #39, opened 2022-08-05
Feature/tooltip

thispastwinter:
## Description

A simple `<Tooltip />` component for displaying non-interactive descriptive text that is anchored to its child element. Supports keyboard accessibility.

[See it in the wild](https://www.justinklaas.com/projects/1) (hover over the tech icons)

### 🔗 Reference

- [Issues #28](https://github.com/cremalab/crema-components/issues/28)

### 📋 Acceptance Criteria

- [ ] CRITERION_HERE

### 🖼 Demo

https://user-images.githubusercontent.com/46500734/183151137-492a3cb6-d929-4276-ab26-3e2bc469bf6a.mov

### 👩‍🔬 Test Instructions

1. run `npm run test:playground`
2. ensure tooltip behaves as intended

## 🔎 Reviewer Checklist

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>

### DIFF
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -1,0 +1,14 @@
+import decoratorCentered from "@storybook/addon-centered"
+import { Tooltip } from "./Tooltip"
+
+/**
+ * See Storybook Docs: Writing Stories
+ * https://storybook.js.org/docs/basics/writing-stories/
+ */
+
+export default {
+  title: "Tooltip",
+  decorators: [decoratorCentered],
+}
+
+export const Example = () => <Tooltip />
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -1,15 +1,12 @@
 import decoratedCenter from "@storybook/addon-centered"
 import { ComponentStory } from "@storybook/react"
+import { placements } from "./placements"
 import { Tooltip } from "./Tooltip"
 
 /**
  * See Storybook Docs: Writing Stories
  * https://storybook.js.org/docs/basics/writing-stories/
  */
-
-const basePlacement = ["top", "bottom", "left", "right", "auto"]
-const startPlacement = basePlacement.map((placement) => `${placement}-start`)
-const endPlacement = basePlacement.map((placement) => `${placement}-end`)
 
 export default {
   title: "Components/Tooltip",
@@ -18,7 +15,7 @@ export default {
     placement: {
       control: {
         type: "select",
-        options: [...basePlacement, ...startPlacement, ...endPlacement],
+        options: placements,
       },
     },
   },
@@ -60,30 +57,26 @@ WithoutArrow.args = {
 
 export const WithAdditionalContent: ComponentStory<typeof Tooltip> = (args) => {
   return (
-    <div style={{ display: "flex", alignItems: "center" }}>
-      <div>
-        <p>
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque
-          in ipsum id orci porta dapibus.
-        </p>
-      </div>
-      <div>
-        <p>Some text above.</p>
-        <Tooltip {...args} />
-        <p>Some text below.</p>
-      </div>
-      <div>
-        <p>
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque
-          in ipsum id orci porta dapibus.
-        </p>
-      </div>
+    <div
+      style={{
+        lineHeight: "24px",
+        width: 400,
+      }}
+    >
+      <p>
+        This is a tooltip that lives within a <Tooltip {...args} />. Tooltips
+        can be very handy for communicating additional information within a body
+        of text :).
+      </p>
     </div>
   )
 }
 
 WithAdditionalContent.args = {
   ...Default.args,
+  children: <b>paragraph</b>,
+  label:
+    "a distinct section of a piece of writing, usually dealing with a single theme and indicated by a new line, indentation, or numbering.",
   showArrow: true,
   placement: "bottom",
   alwaysShow: false,
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -39,6 +39,8 @@ Default.args = {
   showArrow: false,
   enterDelay: 0,
   exitDelay: 0,
+  hideOnClick: false,
+  alwaysShow: false,
 }
 
 export const WithArrow = Template.bind({})
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -1,4 +1,5 @@
-import decoratorCentered from "@storybook/addon-centered"
+import decoratedCenter from "@storybook/addon-centered"
+import { ComponentStory } from "@storybook/react"
 import { Tooltip } from "./Tooltip"
 
 /**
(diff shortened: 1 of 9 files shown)


luiz-chagas:
Thanks for working on this component! I was running the code and put a tooltip inside one of the Tabs in the App component, the result seems a bit off 🤔.

In the picture I'm hovering "Hello" in the Tab 3 and "World" is popping up at the very top
Screen Shot 2022-08-05 at 3 32 14 PM

Code:
```tsx
      <Tabs currentTab={currentTab} onTabChange={setCurrentTab}>
        <Tab id="1" label="Tab 1">
          Content 1
        </Tab>
        <Tab id="2" label="Tab 2">
          Content 2
        </Tab>
        <Tab id="3" label="Tab 3">
          <Tooltip label="world">Hello</Tooltip>
        </Tab>
      </Tabs>
```

thispastwinter:
> Thanks for working on this component! I was running the code and put a tooltip inside one of the Tabs in the App component, the result seems a bit off 🤔.
> 
> 
> 
> In the picture I'm hovering "Hello" in the Tab 3 and "World" is popping up at the very top
> 
> Screen Shot 2022-08-05 at 3 32 14 PM
> 
> 
> 
> Code:
> 
> ```tsx
> 
>       <Tabs currentTab={currentTab} onTabChange={setCurrentTab}>
> 
>         <Tab id="1" label="Tab 1">
> 
>           Content 1
> 
>         </Tab>
> 
>         <Tab id="2" label="Tab 2">
> 
>           Content 2
> 
>         </Tab>
> 
>         <Tab id="3" label="Tab 3">
> 
>           <Tooltip label="world">Hello</Tooltip>
> 
>         </Tab>
> 
>       </Tabs>
> 
> ```
> 
> 

Yes, I just noticed something similar in another project I dropped this in to test in the "real world." I have an idea of how to fix it. Good catch. It's currently wrapped in a fragment so I believe it ends up inheriting the layout of its parent. Might be a use case for a portal.

thispastwinter:
Alright @luiz-chagas I've resolved this placement issue. It was actually because I failed to follow the documentation and wasn't setting the state for each element within the ref callback. Each element needs to be set to state versus a ref to ensure the positioning is always accurate. I tested it against the example above with success 😄 .

thispastwinter:
Tests are failing. Looking into it.

src/components/Tooltip/placements.ts
```diff
@@ -0,0 +1,13 @@
...
+const startPlacement = basePlacement.map((placement) => `${placement}-start`)
+const endPlacement = basePlacement.map((placement) => `${placement}-end`)
+
+// while @popperjs/core has its own placements array, in some cases the export is undefined, specifically within a testing environment
+
+export const placements = [
+  ...basePlacement,
+  ...startPlacement,
+  ...endPlacement,
+] as Placement[]
```
roblafeve: This is minor, but I'd suggest using a type annotation rather than an assertion (`as`) so that we can be sure that our assignment really does conform to `Placement[]`. As it is, if you change `...-end` to `...-eeend` on line 5, no error is raised.

```suggestion
import { Placement } from "@popperjs/core"

const basePlacement = ["top", "bottom", "left", "right", "auto"] as const
const startPlacement = basePlacement.map((placement) => `${placement}-start` as const)
const endPlacement = basePlacement.map((placement) => `${placement}-end` as const)

// while @popperjs/core has its own placements array, in some cases the export is undefined, specifically within a testing environment

export const placements: Placement[] = [
  ...basePlacement,
  ...startPlacement,
  ...endPlacement,
]
```

roblafeve (commented):
I haven't gotten a chance to do a full review, but I noticed something in my first pass that might be helpful.

src/components/Tooltip/placements.ts
thispastwinter: Nice, good suggestion.

thispastwinter: I didn't realize you could use const within the map itself so that's super handy.

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+    act(() => {
+      jest.advanceTimersByTime(exitDelay)
+    })
+
+    expect(screen.getByText(label)).not.toBeVisible()
+  })
+  it("supports aria-describedby", async () => {
+    // Arrange
+    jest.useRealTimers()
```
luiz-chagas: This looks a bit odd, one test is calling `useRealTimers` _probably_ because the previous test is using `useFakeTimers`, there's now an implicit relationship between tests that one test _has_ to come after the other.

Would this be better if the test calling `useFakeTimers` also called `useRealTimers` at the end like some sort of "cleanup"? And then other tests might not have to care whether or not `fakeTimers` are being used.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,166 @@
...
+  useState,
+} from "react"
+import { usePopper } from "react-popper"
+import { Transition, TransitionStatus } from "react-transition-group"
+import classes from "./Tooltip.module.css"
+
+interface TooltipProps {
+  children: ReactNode
+  label: string
+  ariaDescribedBy?: string
```
luiz-chagas: Would it be worth extending an extensive list of aria properties or is `described by` the only one that makes sense here?

thispastwinter: I was going off of this documentation here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

Feel free to poke around and see if I'm missing anything. Happy to expose more aria properties if we feel that is best.

src/components/Tooltip/Tooltip.test.tsx
thispastwinter: Yeah you are right. It makes sense to contain the timer logic within that particular case. A cleanup will be perfect 😄

src/components/Tooltip/Tooltip.tsx
luiz-chagas: It looks like `described by` might be all that's needed :+1:

My question comes from the interfaces that we have been using for components here, more often than not we just expose one / a few aria attributes but I wonder if we could extend `AriaAttributes` and just give the user the full list of aria attributes. 

But then users would need to know which ones to use, if we explicitly expose them that's kind of a hint of what attributes they should be using.

All this rambling to come to the conclusion that it's probably best to expose a subset of aria attributes instead of the whole thing.

albatrocity: Could we `<Pick>` the relevant attributes from `AriaAttributes` instead of retyping them?

thispastwinter: I'm good with that. Like this? `interface TooltipProps extends Pick<AriaAttributes, "aria-describedby">`

thispastwinter: then in props `  ...aria`

and finally implemented like so ` aria-describedby={aria["aria-describedby"]}`

thispastwinter: Or I guess at that point we can just spread the relevant attributes.

albatrocity: @thispastwinter yeah I'd think so. Not totally clear on how the implementation would look like, just a thought.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { Tooltip } from "./Tooltip"
+
+/**
+ * See Storybook Docs: Writing Stories
+ * https://storybook.js.org/docs/basics/writing-stories/
+ */
+
+export default {
+  title: "Components/Tooltip",
+  decorators: [decoratedCenter],
```
belv2c: If we remove `[decoratedCenter]`, it should fix the issue with the component scrolling on the Storybook docs page

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,169 @@
...
+      <span
+        data-testid="tooltip_span"
+        role={hideOnClick ? "button" : "none"}
+        tabIndex={0}
+        onKeyDown={handleKeyDown}
+        onClick={handleClick}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+        ref={setAnchorElement}
+        // ensures screen reader can read out tootlip content without displaying it
```
belv2c: [nit]: I'm not sure if we need this screen reader comment right here? I feel like `aria-label` is a commonly used aria attribute.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+WithoutArrow.args = {
+  ...Default.args,
+  alwaysShow: true,
+}
+
+export const WithAdditionalContent: ComponentStory<typeof Tooltip> = (args) => {
+  return (
+    <div
+      style={{
+        lineHeight: "24px",
```
roblafeve, line 74: ⛏️ I was surprised to see the line-height used for the paragraph affecting the tooltip's line-height. Would it make sense for `.box` to have its own default line-height?

Screen Shot 2022-08-15 at 4 14 41 PM

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,168 @@
...
+        onClick={handleClick}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+        ref={setAnchorElement}
+        aria-label={label}
+        {...aria}
+      >
+        {children}
+      </span>
+      <Transition in={isOpen || alwaysShow} timeout={animationDuration}>
```
roblafeve: ❔I notice that the tooltip content is always rendered even when hidden. Is this intended?

roblafeve (commented):
Solid work! There were just a few things I noticed...

roblafeve (approved):
👏

PR #41, opened 2022-08-12
Feature/search input

thispastwinter:
## Description

This PR adds a minimal `<SearchInput />` offering a few helpful callbacks including a debounced callback. The simplest implementation would be to simply use the onDebounce callback and grab the value anytime its updated based on the debounceDelay (default of 300ms). If you wish to include a search button, simply supply a onSearchClick callback which will return the current value when the search button is clicked.

### 🔗 Reference

- [LINK_NAME](LINK_URL)

### 📋 Acceptance Criteria

- [ ] CRITERION_HERE

### 🖼 Demo
Screen Shot 2022-08-15 at 4 56 50 PM

### 👩‍🔬 Test Instructions

1. TEST_STEP_HERE

## 🔎 Reviewer Checklist

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>

### DIFF
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -31,3 +31,4 @@ The following props are unique to the `<SearchInput />`. It also implements all 
 | `clearIcon`        | ReactNode | no       |                                                                 | ❎      |
 | `debounceDelay`    | number    | no       | the debounce delay in ms                                        | 300     |
 | `onSearch`         | function  | no       | A debounced callback that returns the current search term value |         |
+| `initialValue`     | string    | no       | the initial value of the input                                  |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -1,6 +1,6 @@
 # `<SearchInput />`
 
-DESCRIPTION_HERE
+A minimal search input offering a few helpful callbacks including a debounced callback. Can be used as a controlled or uncontrolled input. For full control, supply your own onChange and value props. The simplest implementation would be to simply use the onDebounce callback and grab the value anytime its updated based on the debounceDelay (default of 300ms). If you wish to include a search button, simply supply a onSearchClick callback which will return the current value when the search button is clicked.
 
 ## Directory Structure
 
@@ -9,6 +9,19 @@ DESCRIPTION_HERE
 - `SearchInput.tsx`: Component code
 - `index.ts`: Component export
 - `README.md`: Component documentation (hey, that's me!)
+
+# Example Usage
+
+```tsx
+function MyComponent() {
+
+  const [value, setValue] = useState("")
+
+  return (
+    <SearchInput onDebounce={(text) => doSomethingWithDebouncedText(text)} />
+  )
+}
+```
 
 ## API
 
@@ -20,4 +33,4 @@ DESCRIPTION_HERE
 | `onDebounce`    | string                  | no       | A debounced callback that returns the current search term value |         |
 | `onSearchClick` | string                  | no       | A callback that returns the current search term value           |         |
 | `value`         | string                  | no       | The current input value                                         |         |
-| `inputProps`    | ComponentProps<"input"> | no       | excludes `name` and `placeholder`                               |         |
+| `...rest`       | ComponentProps<"input"> | no       |                                                                 |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -1,0 +1,11 @@
+# `<SearchInput />`
+
+DESCRIPTION_HERE
+
+## Directory Structure
+
+- `SearchInput.stories.tsx`: Component stories (`npm run test:playground`)
+- `SearchInput.test.tsx`: Component tests (`npm run test:unit`)
+- `SearchInput.tsx`: Component code
+- `index.ts`: Component export
+- `README.md`: Component documentation (hey, that's me!)
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -14,9 +14,7 @@ A minimal search input offering a few helpful callbacks including a debounced ca
 
 ```tsx
 function MyComponent() {
-  return (
-    <SearchInput onSearch={(text) => doSomethingWithOnSearchText(text)} />
-  )
+  return <SearchInput onSearch={(text) => doSomethingWithOnSearchText(text)} />
 }
 ```
 
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -16,7 +16,8 @@ DESCRIPTION_HERE
 | --------------- | ----------------------- | -------- | --------------------------------------------------------------- | ------- |
 | `searchIcon`    | ReactNode               | no       |                                                                 | 🔍      |
 | `clearIcon`     | ReactNode               | no       |                                                                 | ❎      |
-| `debounceDelay` | number                  | no       | the debounce delay in ms                                        |         |
+| `debounceDelay` | number                  | no       | the debounce delay in ms                                        | 300     |
 | `onDebounce`    | string                  | no       | A debounced callback that returns the current search term value |         |
 | `onSearchClick` | string                  | no       | A callback that returns the current search term value           |         |
-| `...rest`       | ComponentProps<"input"> |          |                                                                 |         |
+| `value`         | string                  | no       | The current input value                                         |         |
+| `inputProps`    | ComponentProps<"input"> | no       | excludes `name` and `placeholder`                               |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -10,13 +10,13 @@ DESCRIPTION_HERE
 - `index.ts`: Component export
 - `README.md`: Component documentation (hey, that's me!)
 
-## API 
+## API
 
-| Prop            | Type                    | Required | Description            | Default |
-| --------------- | ----------------------- | -------- | ---------------------- | ------- |
-| `searchIcon`     | ReactNode               | no       |                        |    🔍     |
-| `clearIcon`       | ReactNode               | no       |           |    ❎     |
-| `debounceDelay`       | number               | no       | the debounce delay in ms          |         |
-| `onDebounce`    | string                  | no       | A debounced callback that returns the current search term value          |         |
-| `onSearchClick` | string                  | no       | A callback that returns the current search term value          |         |
-| `...rest`       | ComponentProps<"input"> |          |                        |         |
+| Prop            | Type                    | Required | Description                                                     | Default |
+| --------------- | ----------------------- | -------- | --------------------------------------------------------------- | ------- |
+| `searchIcon`    | ReactNode               | no       |                                                                 | 🔍      |
+| `clearIcon`     | ReactNode               | no       |                                                                 | ❎      |
+| `debounceDelay` | number                  | no       | the debounce delay in ms                                        |         |
+| `onDebounce`    | string                  | no       | A debounced callback that returns the current search term value |         |
+| `onSearchClick` | string                  | no       | A callback that returns the current search term value           |         |
+| `...rest`       | ComponentProps<"input"> |          |                                                                 |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -20,7 +20,7 @@ function MyComponent() {
 
 ## API
 
-The following props are unique to the `<SearchInput />`. It also implements all props belonging to the native `<input />` less `onChange` and `type`. Passing a value is not necessary but may be relevant when needing to pass in a value from url parameters or other scenarios where a value needs persistence. 
+The following props are unique to the `<SearchInput />`. It also implements all props belonging to the native `<input />` less `onChange` and `type`. Passing a value is not necessary but may be relevant when needing to pass in a value from url parameters or other scenarios where a value needs persistence.
 
 | Prop               | Type      | Required | Description                                                     | Default |
 | ------------------ | --------- | -------- | --------------------------------------------------------------- | ------- |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -14,10 +14,8 @@ A minimal search input offering a few helpful callbacks including a debounced ca
 
 ```tsx
 function MyComponent() {
-  const [value, setValue] = useState("")
-
   return (
-    <SearchInput onDebounce={(text) => doSomethingWithDebouncedText(text)} />
+    <SearchInput onSearch={(text) => doSomethingWithOnSearchText(text)} />
   )
 }
 ```
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -1,6 +1,6 @@
 # `<SearchInput />`
 
-A minimal search input offering a few helpful callbacks including a debounced callback. Can be used as a controlled or uncontrolled input. For full control, supply your own onChange and value props. The simplest implementation would be to simply use the onDebounce callback and grab the value anytime its updated based on the debounceDelay (default of 300ms). If you wish to include a search button, simply supply a onSearchClick callback which will return the current value when the search button is clicked.
+A minimal search input offering a few helpful callbacks including a debounced callback. The simplest implementation would be to simply use the onDebounce callback and grab the value anytime its updated based on the debounceDelay (default of 300ms). If you wish to include a search button, simply supply a onSearchClick callback which will return the current value when the search button is clicked.
 
 ## Directory Structure
 
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -20,6 +20,8 @@ function MyComponent() {
 
 ## API
 
+The following props are unique to the `<SearchInput />`. It also implements all props belonging to the native `<input />` less `onChange` and `type`. Passing a value is not necessary but may be relevant when needing to pass in a value from url parameters or other scenarios where a value needs persistence. 
+
 | Prop               | Type      | Required | Description                                                     | Default |
 | ------------------ | --------- | -------- | --------------------------------------------------------------- | ------- |
 | `name`             | string    | no       |                                                                 |         |
@@ -29,6 +31,3 @@ function MyComponent() {
 | `clearIcon`        | ReactNode | no       |                                                                 | ❎      |
 | `debounceDelay`    | number    | no       | the debounce delay in ms                                        | 300     |
 | `onSearch`         | function  | no       | A debounced callback that returns the current search term value |         |
-| `onBlur`           | function  | no       | A callback triggered when the input is blurred                  |         |
-| `onFocus`          | function  | no       | A callback triggered when the input is focused                  |         |
-| `value`            | string    | no       | The initial value to display                                    |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -24,12 +24,15 @@ function MyComponent() {
 
 ## API
 
-| Prop            | Type                    | Required | Description                                                     | Default |
-| --------------- | ----------------------- | -------- | --------------------------------------------------------------- | ------- |
-| `searchIcon`    | ReactNode               | no       |                                                                 | 🔍      |
-| `clearIcon`     | ReactNode               | no       |                                                                 | ❎      |
-| `debounceDelay` | number                  | no       | the debounce delay in ms                                        | 300     |
-| `onDebounce`    | string                  | no       | A debounced callback that returns the current search term value |         |
-| `onSearchClick` | string                  | no       | A callback that returns the current search term value           |         |
-| `value`         | string                  | no       | The current input value                                         |         |
-| `...rest`       | ComponentProps<"input"> | no       |                                                                 |         |
+| Prop            | Type      | Required | Description                                                     | Default |
+| --------------- | --------- | -------- | --------------------------------------------------------------- | ------- |
+| `name`          | string    | no       |                                                                 |         |
+| `placeholder`   | string    | no       |                                                                 |         |
+| `searchIcon`    | ReactNode | no       |                                                                 | 🔍      |
+| `clearIcon`     | ReactNode | no       |                                                                 | ❎      |
+| `debounceDelay` | number    | no       | the debounce delay in ms                                        | 300     |
+| `onSearch`      | function  | no       | A debounced callback that returns the current search term value |         |
+| `onSearchClick` | function  | no       | A callback that returns the current search term value           |         |
+| `onBlur`        | function  | no       | A callback triggered when the input is blurred                  |         |
+| `onFocus`       | function  | no       | A callback triggered when the input is focused                  |         |
+| `initialValue`  | string    | no       | The initial value to display                                    |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -9,3 +9,14 @@ DESCRIPTION_HERE
 - `SearchInput.tsx`: Component code
 - `index.ts`: Component export
 - `README.md`: Component documentation (hey, that's me!)
+
+## API 
+
+| Prop            | Type                    | Required | Description            | Default |
+| --------------- | ----------------------- | -------- | ---------------------- | ------- |
+| `searchIcon`     | ReactNode               | no       |                        |    🔍     |
+| `clearIcon`       | ReactNode               | no       |           |    ❎     |
+| `debounceDelay`       | number               | no       | the debounce delay in ms          |         |
+| `onDebounce`    | string                  | no       | A debounced callback that returns the current search term value          |         |
+| `onSearchClick` | string                  | no       | A callback that returns the current search term value          |         |
+| `...rest`       | ComponentProps<"input"> |          |                        |         |
--- a/src/components/SearchInput/README.md
+++ b/src/components/SearchInput/README.md
@@ -24,15 +24,15 @@ function MyComponent() {
 
 ## API
 
-| Prop            | Type      | Required | Description                                                     | Default |
-| --------------- | --------- | -------- | --------------------------------------------------------------- | ------- |
-| `name`          | string    | no       |                                                                 |         |
-| `placeholder`   | string    | no       |                                                                 |         |
-| `searchIcon`    | ReactNode | no       |                                                                 | 🔍      |
-| `clearIcon`     | ReactNode | no       |                                                                 | ❎      |
-| `debounceDelay` | number    | no       | the debounce delay in ms                                        | 300     |
-| `onSearch`      | function  | no       | A debounced callback that returns the current search term value |         |
-| `onSearchClick` | function  | no       | A callback that returns the current search term value           |         |
-| `onBlur`        | function  | no       | A callback triggered when the input is blurred                  |         |
-| `onFocus`       | function  | no       | A callback triggered when the input is focused                  |         |
-| `initialValue`  | string    | no       | The initial value to display                                    |         |
+| Prop               | Type      | Required | Description                                                     | Default |
+| ------------------ | --------- | -------- | --------------------------------------------------------------- | ------- |
+| `name`             | string    | no       |                                                                 |         |
+| `placeholder`      | string    | no       |                                                                 |         |
+| `showSearchButton` | boolean   | no       | whether to show the search button or not                        |         |
+| `searchIcon`       | ReactNode | no       |                                                                 | 🔍      |
+| `clearIcon`        | ReactNode | no       |                                                                 | ❎      |
+| `debounceDelay`    | number    | no       | the debounce delay in ms                                        | 300     |
+| `onSearch`         | function  | no       | A debounced callback that returns the current search term value |         |
+| `onBlur`           | function  | no       | A callback triggered when the input is blurred                  |         |
+| `onFocus`          | function  | no       | A callback triggered when the input is focused                  |         |
+| `initialValue`     | string    | no       | The initial value to display                                    |         |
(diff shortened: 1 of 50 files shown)


thispastwinter:
Just fyi that I have several tweaks I'll be shipping out today to make sure this better represents the expected behaviors of a search input.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+import styles from "./SearchInput.module.css"
+
+interface SearchInputProps extends ComponentProps<"input"> {
+  searchIcon?: ReactNode
+  clearIcon?: ReactNode
+  onDebounce?: (searchTerm: string) => void
+  debounceDelay?: number
+  onSearchClick?: (searchTerm: string) => void
+  value?: string
+}
```
luiz-chagas: `onDebounce` feels like Leaky Abstraction, so does `onSearchClick`. 

Debounce is a technique for dealing with event dispatching, preventing multiple events to be dispatched in a short amount of time.

Whether by user clicking or a debounce timing out, the important thing about this component is that it triggers a search action with some `(searchTerm: string)` (notice how it shows up twice in the interface), so it would be expected in my opinion to see a `onSearch` callback prop that _could_ be triggered by click or debounce, and the beauty of it is that the consumer of this component doesn't have to know/worry about clicks and debounces.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [searchTerm, setSearchTerm] = useState(value)
+  const [hideCancelButton, setHideCancelButton] = useState(true)
+
+  useEffect(() => {
+    const timer = setTimeout(() => onDebounce?.(searchTerm), debounceDelay)
+    return () => {
+      clearTimeout(timer)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [searchTerm, debounceDelay])
```
luiz-chagas: I haven't properly tested this but it looks like `onDebounce` gets called automatically after the timeout in the initial mount, we might want to see if there's a `searchTerm` before triggering the action

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  placeholder,
+  searchIcon,
+  clearIcon,
+  onDebounce,
+  debounceDelay = 300,
+  onSearchClick,
+  value = "",
+  ...inputProps
+}: SearchInputProps) {
+  const [searchTerm, setSearchTerm] = useState(value)
```
luiz-chagas: Saving props to state usually opens a can of worms when the props (`value`) change on the fly

src/components/SearchInput/SearchInput.tsx
thispastwinter: Yeah, that's a great callout. It certainly does.

src/components/SearchInput/SearchInput.tsx
thispastwinter: Yeah, I originally had it set via a useEffect. I think I was aiming for less code, but I totally see what you mean given the value is highly dynamic. I also really went back and forth on whether this component should control any kind of state. I could just as easily include an onClear callback and ensure all state management is contained in the parent.

albatrocity: I think if the intention is to copy/paste this component as a sort of boilerplate/head start, it should be controlled (no local state). An implementor is going to have to use the `onChange` handler to get the value anyway (though I prefer @luiz-chagas suggestion of an `onSearch` handler and abstracting the input away entirely). If I'm passing `value` in, I don't expect anything else to manage it.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+          placeholder={placeholder}
+          onChange={handleChange}
+        />
+        <button
+          // we are ignoring tabbing since 'esc' is the keyboard accessible means of clearing an input
+          tabIndex={-1}
+          aria-label="click icon to clear search"
+          hidden={hideCancelButton}
+          className={styles.clearIcon}
+          onMouseDown={handleReset}
```
albatrocity, line 103: I tend to think of these type of actions happening `onMouseUp` instead of `onMouseDown`, if only because it gives the user an escape hatch if they want to cancel their action. Is there a precedent for using `onMouseDown` here?

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+      clearTimeout(timer)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [searchTerm, debounceDelay])
+
+  const handleReset = (e: MouseEvent<HTMLButtonElement>) => {
+    // we don't want the field to blur when a user clears the input
+    e.preventDefault()
+    setSearchTerm("")
+    setHideCancelButton(true)
```
albatrocity: As a controlled component, the parent component will never know when the value has been reset. You'll need to call `inputProps.onChange?.(e)`. Also, `onDebounce` is called with wacky arguments after clearing.

https://user-images.githubusercontent.com/5196/184888600-00a63d53-72b8-4541-a555-73cb5e11a8ff.mov

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  searchIcon,
+  clearIcon,
+  onDebounce,
+  debounceDelay = 300,
+  onSearchClick,
+  value = "",
+  ...inputProps
+}: SearchInputProps) {
+  const [searchTerm, setSearchTerm] = useState(value)
+  const [hideCancelButton, setHideCancelButton] = useState(true)
```
albatrocity: nit ⛏️ : avoid naming boolean state as a verb. When first reading through this I thought `hideCancelButton` was a function. `setHideCancelButton` is weird to read. Maybe something like `cancelHidden` and `setCancelHidden`?

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        />
+        <button
+          // we are ignoring tabbing since 'esc' is the keyboard accessible means of clearing an input
+          tabIndex={-1}
+          aria-label="click icon to clear search"
+          hidden={hideCancelButton}
+          className={styles.clearIcon}
+          onMouseDown={handleReset}
+        >
+          {clearIcon || <span>❎</span>}
```
albatrocity: I like the ability to pass in a custom clear and search icons 👍

src/components/SearchInput/SearchInput.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    const input = screen.getByLabelText(label)
+
+    await user.type(input, text)
+
+    act(() => {
+      jest.advanceTimersByTime(300)
+    })
+
+    // Assert
+    await waitFor(() => expect(onDebounce).toBeCalled())
```
albatrocity: nit ⛏️ : this test would pass if the default delay was any number less than 300. You could add an expect before the `waitFor` to assert that it hasn't been fired before the debounce time.

albatrocity (changes requested):
Added some nitpicks and some questions about concern. Only real must-change is `onDebounce` not firing as expected when clearing the input.

src/components/SearchInput/SearchInput.tsx
thispastwinter: I'll double back and verify, but I'm pretty certain I had to take this route in order prevent the input from blurring when the clear button is pressed. Otherwise, the onClick and preventDefault wasn't working as expected. Definitely open to learning something here. Maybe I'm not understanding the problem fully.

src/components/SearchInput/SearchInput.tsx
thispastwinter: Originally the intention was to have it be uncontrolled. You would use the `onDebounce` or @luiz-chagas suggested better name of `onSearch` and that was that. I then back peddled some and overcomplicated it by trying to marry some form of controlled/uncontrolled but it clearly became more fragile. If we want it to be fully controlled, I think we have an onChange, value, and onClear that are required. Then it is left to the user to leverage them in a way that makes sense.

thispastwinter: I'm leaning that way, so at that point this is just really a glorified input that sort of represents a search bar....which if I'm being honest, I probably wouldn't reach for. That said, I also don't want to create something that is fragile and overly complicated.

thispastwinter: If we want to further strip it down, instead of a left and right icon with search related names, it could just be a left/right element. At that point you are responsible for any sort of events you would want to assign to those elements. 

Accept now it's just an input with a start adornment and end adornment. At that point it might as well be part of a larger composite like our existing TextInput, which is more like a TextField as it has helper text and a label. However, we are building everything in isolation so everything just gets recreated which just feels wrong. 

That said these thoughts are really just part of a larger conversation and reflect my strong desire to build something representing more of a library than a copy/paste repo.

albatrocity: Thinking through this more, the case for a controlled, debounced component is more complicated when trying to keep it in sync with outside changes. I had to write this scenario down and think through the implementation to make much sense of either approach:

I have a search page at `/search` that I want to use a Search Input component on. Since I want to support deep-linking and browser history as people use the search page, the search query that I want send to my API is represented in the URL as `search?q=search%20term`.

## Expectations
* search results and search query are updated when the URL changes (from history event)
* URL is updated when `onSearch` is called (after debounced user input)

### Uncontrolled
* container passes value from `?q=` to the `value` prop
* container performs a route update on `onSearch`
* container does not pass `onChange` callback to Search Input
* Search Input needs to watch its own `value` prop with `useEffect` ([as you mentioned](https://github.com/cremalab/crema-components/pull/41/files#r946310242)) to `setSearchTerm` and call the debounce handler in order to keep the search term in sync with the URL param

### Controlled
* container manages search input's _actual current value_ in state somewhere, passes it to the `value` prop, updates it when Search Input's `onChange` event fires
* container performs a route update on `onSearch`
* container watches value from `?q=` and updates value sent to Search Input's `value` when it changes

I'm not clear on a pattern that would simplify this other than making it _feel_ controlled but using the "uncontrolled" approach above. 

🤷

thispastwinter: Had some thoughts over lunch. Let me play around. I'll ping you once I've implemented them :)

src/components/SearchInput/SearchInput.tsx
thispastwinter: Okay, I figured out what's happening here. It looks like the hidden attribute along with other methods of hiding elements disables pointer-events...which is the desired behavior...but it seems that the boolean hides the element before onClick has a chance to fire which is why I ended up with onMouseDown. I'm going to see if there is a better strategy here with styling where I can still leverage onClick.

thispastwinter: So in this case, it seems onMouseDown might be the best strategy to ensure that handleReset is fired before the element is hidden. If you think of a better way, let me know!

thispastwinter: setting `opacity: 0; pointer-events: none`, `visibility: hidden;`, and `display: none; pointer-events: none`; all seem to remove pointer-events before onClick can be fired. Disabling the element has the same effect as well.

thispastwinter:
@luiz-chagas @albatrocity I went full on uncontrolled. Curious to hear your thoughts on the changes.

src/components/SearchInput/SearchInput.tsx
thispastwinter: @luiz-chagas so I moved forward with this advice. However, I think it loses the convenience of avoiding the callback being fired until the button is clicked. Trying to think of the best strategy here. Do we maybe just not debounce when the `showSearchButton` boolean is `true`?

thispastwinter: My other thought was returning the event type in the callback so the developer could programmatically decided how to handle each type.

luiz-chagas:
> @luiz-chagas @albatrocity I went full on uncontrolled. Curious to hear your thoughts on the changes.

You were always on controlled land, I think the only way to go uncontrolled is by using refs and not setting `value` on inputs.
That said, it's looking good!

src/components/SearchInput/SearchInput.tsx
thispastwinter: I'm leaning towards the former. The latter seems unnecessarily complex and at that point you might as well have two different callbacks.

thispastwinter:
> e only way to go uncontrolled is

Sure, I guess what I'm distinguishing is where the state is managed. In this case, it's managed fully in the component itself. That's a great distinction though.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  useEffect(() => {
+    let timer: NodeJS.Timeout | undefined
+    if (searchTerm) {
+      timer = setTimeout(() => onSearch?.(searchTerm), debounceDelay)
+    }
+    return () => {
+      if (timer) clearTimeout(timer)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
luiz-chagas: This looks a bit suspicious, if the intent is avoid the effect in case the callback changes (because it could be a new function created on the render of the parent component) I think the best practice is to save it to a `ref`

thispastwinter: Yeah, that's a good call. Not sure why I went that route given that I've used a ref everywhere else in this repo.

src/components/SearchInput/SearchInput.tsx
albatrocity: seems like something that could be refactored later on if we find a solution. `onMouseDown` it is! ✅

src/components/SearchInput/SearchInput.tsx
```diff
@@ -1,41 +1,55 @@
...
-  const [searchTerm, setSearchTerm] = useState(value)
-  const [hideCancelButton, setHideCancelButton] = useState(true)
+  const [searchTerm, setSearchTerm] = useState("")
+  const [cancelButtonHidden, setCancelButtonHidden] = useState(true)
+  const timer = useRef<NodeJS.Timeout>()
 
   useEffect(() => {
-    const timer = setTimeout(() => onDebounce?.(searchTerm), debounceDelay)
+    setSearchTerm(initialValue)
+  }, [initialValue])
```
albatrocity, line 50: :pick: the prop name `initialValue` makes me think that if it changes there will be no effect inside the component's state. Maybe going back to `value` here could be okay? 😬

albatrocity (dismissed):

src/components/SearchInput/SearchInput.tsx
albatrocity: I'm thinking of `initialValues` in [Formik](https://formik.org/docs/api/formik#initialvalues-values), which if changed do not update the current value of the form.

thispastwinter: Yeah, that makes sense. I was trying to get away from the common pattern of value and onChange, but your right, it isn't accurate since when that value changes, the state updates.

thispastwinter:
Haha. GitHub making me look like a jerk. Anyway, @albatrocity I changed initialValue back to value.

src/components/SearchInput/README.md
```diff
@@ -0,0 +1,38 @@
...
+- `README.md`: Component documentation (hey, that's me!)
+
+# Example Usage
+
+```tsx
+function MyComponent() {
+  const [value, setValue] = useState("")
+
+  return (
+    <SearchInput onDebounce={(text) => doSomethingWithDebouncedText(text)} />
```
roblafeve: Looks like we need to update this.

src/components/SearchInput/SearchInput.module.css
```diff
@@ -0,0 +1,53 @@
...
+  outline: none;
+  border: none;
+  padding: 0 4px;
+  width: 100%;
+}
+
+.input::-webkit-search-cancel-button {
+  visibility: hidden;
+  pointer-events: none;
+}
```
roblafeve: I'm seeing the default cancel button in safari, fwiw:

https://user-images.githubusercontent.com/157195/185189844-c64b98cd-9127-4b2e-8892-b0ae8f6089d5.mov

src/components/SearchInput/SearchInput.test.tsx
```diff
@@ -0,0 +1,187 @@
+import { render, screen } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { act } from "react-dom/test-utils"
+import { SearchInput } from "./SearchInput"
+
+describe("SearchInput", () => {
```
roblafeve, line 8: Looks like we're missing 2 branches (16/18):

Screen Shot 2022-08-17 at 11 09 03 AM

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,122 @@
...
+    setSearchTerm(value)
+  }, [value])
+
+  useEffect(() => {
+    if (searchTerm && !showSearchButton) {
+      timer.current = setTimeout(() => onSearch?.(searchTerm), debounceDelay)
+    }
+    return () => {
+      clearTimeout(timer.current)
+    }
```
roblafeve: Cool to see how easily debounce can be pulled off 👌

roblafeve (changes requested):
This is looking solid! I noticed a few things we might want to fix before merging.

src/components/SearchInput/SearchInput.test.tsx
luiz-chagas: I love that you are calling this out! It's one of the reasons I try to avoid `if/else`s at all costs! They automatically create branching logic that is often a pain to test

src/components/SearchInput/SearchInput.module.css
thispastwinter: Ah, good 👁️ . I'll take care of it!

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+  FocusEvent,
+  MouseEvent,
+  ReactNode,
+  useEffect,
+  useRef,
+  useState,
+} from "react"
+import styles from "./SearchInput.module.css"
+
+interface SearchInputProps extends Pick<AriaAttributes, "aria-label"> {
```
belv2c: Have you considered extending the native `HTMLInputElement` element here? This way we can simplify the interface and get rid of placeholder, onBlur, onFocus, etc.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }, [searchTerm, debounceDelay])
+
+  const handleReset = (e: MouseEvent<HTMLButtonElement>) => {
+    // we don't want the field to blur when a user clears the input
+    e.preventDefault()
+    setSearchTerm("")
+    onSearch?.("")
+    setCancelButtonHidden(true)
+  }
+
```
belv2c, line 58: One thing I noticed is we're missing submit handling via the Enter key. I think we can do this by adding:

```
useKeyPress(["Enter"], () => {
    onSearch?.(searchTerm)
  })
```

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+    setCancelButtonHidden(!searchTerm.length)
+  }
+
+  return (
+    <div className={styles.wrapper}>
+      <div className={styles.container} data-testid="container">
+        <span data-testid="search_icon" className={styles.searchIcon}>
+          {searchIcon || <span>🔍</span>}
+        </span>
+        <input
```
belv2c, line 89: If we don't provide a label for the search field, screenreader users may not know what it's used for. What do you think about adding `ariaLabel` as a required prop as a way to ensure descriptive text?

belv2c (changes requested):
This looks great! I just have a few suggestions

src/components/SearchInput/SearchInput.tsx
thispastwinter: I didn't want to do that as I'm trying to avoid the use of onChange. It honestly felt like it bloated it a bit by passing through all of the input props. I just felt like having the onChange loses some of the benefits gained from just using the onSearch callback.

src/components/SearchInput/SearchInput.tsx
thispastwinter: Ah yeah, good catch.

src/components/SearchInput/SearchInput.tsx
thispastwinter: I was considering that and exposing an `id` prop in order to bind a label to the input.

src/components/SearchInput/SearchInput.tsx
thispastwinter: Do you think this should apply in both cases (i.e. with or without the search button)? I'm tempted to say it should apply conditionally on whether the button is present since the debounced callback is fired off anytime the user stops typing when it isn't. Seems like overkill for onSearch to fire after the timeout and additionally with `Enter`.

belv2c: I think it should apply whether or not the search button is present.

src/components/SearchInput/SearchInput.tsx
thispastwinter: @luiz-chagas @roblafeve @albatrocity Feel free to weigh in on this as well. I can be persuaded.

belv2c: If onChange is the only issue we could omit it from the interface: 
`Omit<HTMLProps<HTMLInputElement>, "onChange">`

src/components/SearchInput/SearchInput.tsx
luiz-chagas: >Seems like overkill for onSearch to fire after the timeout and additionally with Enter.

I would expect the timeout not to fire after `Enter` was pressed

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }, [value])
+
+  useEffect(() => {
+    if (searchTerm && !showSearchButton) {
+      timer.current = setTimeout(() => onSearch?.(searchTerm), debounceDelay)
+    }
+    return () => {
+      clearTimeout(timer.current)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
luiz-chagas: I think I lead you astray with my previous comment, the suspicious thing here is the:
```
// eslint-disable-next-line react-hooks/exhaustive-deps
```
I don't see right away why we would need to disable the dependency check and could end up backfiring.

src/components/SearchInput/SearchInput.tsx
thispastwinter: Fair enough

thispastwinter:
Okay @belv2c I've extended the props interface. I'll address the other stuff after my vacation.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+interface SearchInputProps extends Omit<ComponentProps<"input">, Omitted> {
+  "aria-label": string
+  searchIcon?: ReactNode
+  clearIcon?: ReactNode
+  debounceDelay?: number
+  value?: string
+  onSearch?: (searchTerm: string) => void
+  showSearchButton?: boolean
+}
```
luiz-chagas: This API looks a bit different in that it takes a `value` but does not take a `onChange`. There's a slight conflict of responsibilities, who's controlling the value, this component or the parent?

It looks like this component is controlling the value, _unless_ the parent also wants to change the value, then the parent overrides the value, _unless_ there's an input event on this component, and the cycle goes on.

I think I would expect to see a `value` + `onChange` pair or just an `initialValue` and from there the component takes over that responsibility.

But of course this is mostly a philosophical matter, in practice the parent wouldn't be changing the value of `value` and this would work just fine :+1:

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }, [value])
+
+  useEffect(() => {
+    if (shouldInvokeDebounce(searchTerm, showSearchButton)) {
+      timer.current = setTimeout(() => onSearch?.(searchTerm), debounceDelay)
+    }
+    return () => {
+      clearTimeout(timer.current)
+    }
+  }, [searchTerm, onSearch, showSearchButton, debounceDelay])
```
luiz-chagas: I don't think this implementation of `debounce` translates accurately to the original one in that it changes its behavior from imperative to reactive. `debounce` is associated with an input action that _must_ take place before the outgoing action is triggered, but here, changing `onSearch`, `showSearchButton` or `debounceDelay` could trigger `onSearch` to be called, but the input action part of it is exclusively related to the `searchTerm`. 
So, `searchTerm` should be the only thing that could possibly trigger `onSearch`.

In practice it's unlikely that these other variables would be changed on the fly, so, more philosophy here.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,130 @@
...
+      {showSearchButton && (
+        <div className={styles.searchButton}>
+          <button
+            aria-label="click to search"
+            onClick={() => searchTerm && onSearch?.(searchTerm)}
+          >
+            Search
+          </button>
+        </div>
+      )}
```
luiz-chagas: This button isn't `debounce`d, meaning that a user could come here and hit search many times and it could trigger many API requests. If this component is handling `debounce` on `onChange` events, would it make sense to also debounce `onClick` events to keep the whole concept consistent? Or does this go beyond the original intent? (Same with the `keyPress` on `Enter`)

luiz-chagas (dismissed):
This is looking good 🙌🏼 

I left some thought-provoking thoughts/questions but not looking for any changes.

src/components/SearchInput/SearchInput.tsx
thispastwinter: These are great questions @luiz-chagas. I think you're right. Why wouldn't we just debounce any onSearch event. May not really be necessary to handle the conditional of whether a button is present or not. Let me think on this a bit and maybe rework it. I'll ping you after I have.

src/components/SearchInput/SearchInput.tsx
thispastwinter: Yeah, it's a little weird isn't it. I guess I was hoping to enforce the use of onSearch and originally I had added a prop of initialValue over value, except that @albatrocity pointed out that the value is technically being changed, so initialValue may not totally make sense in this case. However, I'm thinking through this further now and if we aren't using onChange, initialValue still makes sense to me, because the prop is essentially used to just pass in what the value should initially be, and the component from there handles mutating that value. I'm still pretty keen on renaming it here, but I'm curious to see what @albatrocity thinks.

albatrocity: I'm on board with changing it back to `initialValue`. My concern was that if the `initialValue` prop changes, resetting the input's value to `initialValue` might be unexpected based on prop name alone. But in reality it's what I would want if triggered by a route change.

src/components/SearchInput/SearchInput.tsx
thispastwinter: @luiz-chagas I think my most recent pass at a debounce handler is the best case scenario. We are now avoiding useEffect and wrapping our debounce function in useMemo to ensure it's not recreated with each render and we are debouncing every event (i.e. onKeyPress, onChange, and onClick).

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+    </div>
+  )
+}
+
+const shouldDebounceOnChange = (showSearchButton?: boolean) => {
+  return !showSearchButton
+}
```
luiz-chagas, line 124: Looking at this function in isolation, it's not easy to see why it's there.

- First, an easy one, this function is a negation of its input, it could've been a statement with less overhead.

- Second, and most importantly, the name isn't clearly related to its input or logic. There's an implicit connection between the function name and debounce here, but what if we removed debounce from this component, would this function still exist? If so, then this function would need to be renamed, to what? And how does that relate with showing the search button (aka, the input)?

This function (or variable if you make it an statement) is helping to fulfill a much higher-level concept:
> If there is a search button present on the screen, we don't want changes in the search value to automatically trigger an action `onSearch`. And then the opposite logic: if there isn't a search button, changes to the search value _should_ trigger the `onSearch` action.

And this logic is independent of debounce, so the challenge here is to pack all of this ☝🏼 into a single function/variable name.

src/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,140 @@
...
+    return debounce(onSearch, debounceDelay)
+  }, [debounceDelay, onSearch])
+
+  useKeyPress(["Enter"], () => {
+    debouncedOnSearch(searchTerm)
+  })
+
+  useEffect(() => {
+    setSearchTerm(initialValue)
+  }, [initialValue])
```
luiz-chagas, line 50: Beware that if the parent changes the `initialValue` on subsequent renders, it will impact the value of `searchTerm`. So it goes beyond the initial value. Which might be fine depending on how you want this component to behave.

Formik (and other libraries, like redux-form) allow both modes of operation in a similar setup with the boolean `enableReinitialize` attribute. If `true`, changes to the `initialValue` affect the state of the component, if `false`, only the initial value is used, and subsequent changes don't affect the state. You def don't need to do what they did though.

luiz-chagas (approved):
This is looking good 🙌🏼 , I left a couple extra thoughts but no changes required.

roblafeve (approved):
Great work! 

I agree with some of Luiz's recent feedback, but again, probably not worth blocking completion.

belv2c (approved):
Nice work! 🎉

PR #47, opened 2022-08-22
Feature/Avatar

thispastwinter:
## Description

A `<Avatar />` component for rendering a circular container used to display a user's initials or an image of a user.

### 🔗 Reference

- [LINK_NAME](LINK_URL)

### 📋 Acceptance Criteria

- [ ] It should always take in a name
- [ ] Its default is to display a first and last initial (if applicable).
- [ ] It takes an optional img src to display an image instead of initials.
- [ ] It can be rendered in different sizes.

### 🖼 Demo

https://user-images.githubusercontent.com/46500734/186004886-7c89158a-7c7c-4693-837b-76f97f35b436.mov

### 👩‍🔬 Test Instructions

1. TEST_STEP_HERE

## 🔎 Reviewer Checklist

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>

### DIFF
--- a/src/components/Avatar/Avatar.module.css
+++ b/src/components/Avatar/Avatar.module.css
@@ -32,4 +32,3 @@
   width: var(--large);
   height: var(--large);
 }
-
--- a/src/components/Avatar/Avatar.module.css
+++ b/src/components/Avatar/Avatar.module.css
@@ -10,23 +10,20 @@
 }
 
 .container[data-size="small"] {
-  font-size: 14px;
+  font-size: 24px;
   width: 48px;
   height: 48px;
-  padding: 8px;
 }
 
 .container[data-size="medium"] {
-  font-size: 18px;
+  font-size: 32px;
   width: 64px;
   height: 64px;
-  padding: 16px;
 }
 
 .container[data-size="large"] {
-  font-size: 24px;
-  width: 72px;
-  height: 72px;
-  padding: 16px;
+  font-size: 48px;
+  width: 96px;
+  height: 96px;
 }
 
--- a/src/components/Avatar/Avatar.module.css
+++ b/src/components/Avatar/Avatar.module.css
@@ -2,6 +2,7 @@
   --small: 48px;
   --medium: 64px;
   --large: 96px;
+  --ratio: 0.5;
 }
 
 .container {
@@ -16,19 +17,19 @@
 }
 
 .container[data-size="sm"] {
-  font-size: calc(var(--small) * 0.5);
+  font-size: calc(var(--small) * var(--ratio));
   width: var(--small);
   height: var(--small);
 }
 
 .container[data-size="md"] {
-  font-size: calc(var(--medium) * 0.5);
+  font-size: calc(var(--medium) * var(--ratio));
   width: var(--medium);
   height: var(--medium);
 }
 
 .container[data-size="lg"] {
-  font-size: calc(var(--large) * 0.5);
+  font-size: calc(var(--large) * var(--ratio));
   width: var(--large);
   height: var(--large);
 }
--- a/src/components/Avatar/Avatar.module.css
+++ b/src/components/Avatar/Avatar.module.css
@@ -1,0 +1,32 @@
+.container {
+  background: #999999;
+  color: #333333;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: 100%;
+  flex-wrap: nowrap;
+  background-size: cover;
+}
+
+.container[data-size="small"] {
+  font-size: 14px;
+  width: 48px;
+  height: 48px;
+  padding: 8px;
+}
+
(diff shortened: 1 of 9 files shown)


src/components/Avatar/Avatar.module.css
```diff
@@ -0,0 +1,35 @@
+:root {
+  --small: 48px;
+  --medium: 64px;
+  --large: 96px;
+  --ratio: 0.5;
+}
+
+.container {
+  background: #999999;
```
roblafeve: ✋ Since `:root` corresponds to the `html` element, this essentially sets these as global variables. With names as generic as these, I can see there being potential conflicts down the road. What if we scope these to the `.container` element?

```suggestion
.container {
  --small: 48px;
  --medium: 64px;
  --large: 96px;
  --ratio: 0.5;
  background: #999999;
```

src/components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { render, screen } from "@testing-library/react"
+import { Avatar, AvatarSize } from "./Avatar"
```
roblafeve: 💡 What do you think of importing modules from the index? This would have the benefit of guaranteeing that we've actually exposed them as intended. Thoughts? (same for stories)

```suggestion
import { Avatar, AvatarSize } from "."
```
> In this can you may want to split up `Avatar` and `AvatarSize` as you may not intend to expose `AvatarSize`—though I don't see the harm. Could be as useful to the end-user as it was to you here. 🤷

src/components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const initials = getInitials(name)
+
+    // Act
+    render(<Avatar name={name} src="someUrl" />)
+
+    const received = screen.queryByText(initials)
+
+    // Assert
+    expect(received).not.toBeInTheDocument()
+  })
```
roblafeve: 💭 Completely subjective stylistic observation and fuel for conversation...

There is more whitespace within your test blocks than between them. This makes it hard for me to scan this file and discriminate between tests. But this may just be me and my subjective expectations. 🤷

src/components/Avatar/README.md
```diff
@@ -0,0 +1,25 @@
...
+<Avatar name="Crema Components" size="md" src="https://www.someurl.com/someimage.jpg">
+```
+
+## API
+
+| Prop   | Type   | Required | Description      | Default |
+| ------ | ------ | -------- | ---------------- | ------- |
+| `name` | string | yes      | A user's name    |         |
+| `size` | string | no       | the avatar size  | "sm"    |
+| `src`  | string | no       | the image source |         |
```
roblafeve, line 25: I like the columns that you include—I need to update my README's to mirror this. 👍

src/components/Avatar/utils/getInitials/getInitials.ts
```diff
@@ -0,0 +1,20 @@
+const capitalize = (str: string) => str.toUpperCase()
+
+const getFirstChar = (str: string) => str.charAt(0)
+
+const getInitial = (str: string) => {
+  const firstChar = getFirstChar(str)
+  const capitalized = capitalize(firstChar)
+  return capitalized
+}
```
roblafeve: 💭 I get how breaking these down in this way can make it easier to think through a solution, but I wonder if it's necessary to create an abstraction that literally saves zero space and doesn't increase the understandability of the code (I think the indirection decreases it):

```suggestion
const getInitial = (str: string) => {
  return str.charAt(0).toUpperCase()
}
```

I think there are a few other culprits in this file as well.

I'm super guilty of doing this sort of thing all the time, but it's something I'm trying to get better at recognizing.

Would you agree with this assessment, or am I off-base?

roblafeve (changes requested):
Solid work! I only have one item that I think is a show-stopper ✋

The rest are just the ramblings of a madman... 😆

src/components/Avatar/Avatar.module.css
thispastwinter: Yeah, great callout, I wasn't sure whether this would scope or not.

src/components/Avatar/utils/getInitials/getInitials.ts
thispastwinter: No I think you're right. Especially given that they are ultimately just wrapping native methods. I think breaking things down into small functions like this can be helpful when they wrap several lines of code versus a one liner in this case. It's not as if we are testing each individually. We are testing the whole.

src/components/Avatar/Avatar.test.tsx
thispastwinter: Wouldn't this create a circular refernce?

roblafeve (approved):
Great work! ✅

PR #48, opened 2022-08-22
45: Add Table component

### LINKED ISSUE #45: Table
## Proposal

Tables can be a pain. Let's make it as quick and easy to spit out a simple table that handles sorting and selection.

### AC

1. Any data can be mapped to a row
2. Can sorting by column
3. Can select by row
4. Does not support [complicated tables](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#complicated_tables)

### A - Render Children and Props

Pros:

1. Kind of looks like a table...

Cons:

1. We'll still have to pass configuration for header/footer to distinct props which will give this a weird shape
2. Need to pull in custom, or native table elements—you essentially still need to build a table
3. Kind of looks like a table—seems like a shallow module (bad)—doesn't hide much

```tsx
<Table
  data={data}
  renderHeader={() => (...)}
  >
  {({ a, b, c, d }) => (
    <Tr>
      <Td>{a}</Td>
      <Td>{b}</Td>
      <Td>{c}</Td>
      <Td>{d}</Td>
    </Tr>
  )}
</Table>
```

### B - Render Props

Pros:

1. Kind of looks like a table
2. All render props are passed as props—as opposed to using `children`

Cons:

1. Same as Option A

```tsx
<Table
  data={data}
  renderHeader={() => (...)}
  renderRow={({ a, b, c, d }) => (
    <Tr>
      <Td>{a}</Td>
      <Td>{b}</Td>
      <Td>{c}</Td>
      <Td>{d}</Td>
    </Tr>
  )}
/>
```

### C - Config-Based

Pros:

1. No knowledge of table structure is needed (deep module)
2. Column headers and table data values are define once—instead of potentially defining them separately—and potentially mismatching them.

Cons:

1. Lose FULL control—but that's the point. This is the starting point for a useful abstraction that the user could augment to fit their needs.

```tsx
<Table
  data={data}
  columns={[
    { label: null }, // maybe we need a gap 🤷‍♂️
    { label: "A", getValue: (user) => user.a, sortable: true },
    { label: "B", getValue: (user) => user.b },
    { label: "C", getValue: (user) => user.c.toString() },
    { label: "D", getValue: (user) => user.d.e },
  ]}
/>
```
> I might convert `columns` to a single function that has "datum" passed to it

roblafeve:
## Description

Adds a component for rendering a simple table that supports sorting

### 🔗 Reference

closes #45

### 📋 Acceptance Criteria

- [x] Can display any data
- [x] Columns are configureable
- [x] Sorting is supported
- [x] Cell rendering
- [ ] Row selection

### 🖼 Demo

https://user-images.githubusercontent.com/157195/187315746-db75d009-286e-45e2-814f-1a36843d900b.mov

### 👩‍🔬 Test Instructions

1. TBD

## 🔎 Reviewer Checklist

- [ ] Code has been tested and confirmed locally

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>


src/components/Table/types.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: string
+}
+
+export interface ColumnConfig<Datum extends WithID> {
+  label: string | null
+  getValue?: (datum: Datum) => ReactNode
+  sortable?: boolean
+}
+
+export type ColumnConfigs<Datum extends WithID> = ColumnConfig<Datum>[]
```
luiz-chagas: To me it looks like these 2 types belong to the table, I think having them in the Table file would improve readability, as one wouldn't have to jump to a different file to see what the type means.

luiz-chagas: Also, not sure how I feel about exporting collection types (`ColumnConfigs`, a collection of `ColumnConfig`).
It's definitely not wrong, but also not common practice, I don't know why. 🤔 

Is it assumed that making a collection is trivial? Not worth the extra characters in the bundle? Maybe the user wants to name the type something else? `ColumnConfigList`? None of these arguments seem strong enough to me.

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,70 @@
...
+        {columnConfig.label}
+        {columnConfig === sortConfig ? (sortAsc ? "↑" : "↓") : " "}
+      </th>
+    )
+  })
+
+  const rows = data.map((datum) => (
+    <tr key={`tr-${datum.id}`}>
+      {columnConfigs.map((columnConfig, i) => (
+        <td key={`tr-${datum.id}-td-${i}`}>{columnConfig.getValue?.(datum)}</td>
```
luiz-chagas: It looks like there's only valuable output when `getValue` is defined, should it be required? What's the case for it being optional?

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <table>
+      <thead>
+        <tr>{headers}</tr>
+      </thead>
+      <tbody>{rows}</tbody>
+    </table>
+  )
+}
```
luiz-chagas: Nice `return`, very easy to follow

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ColumnConfig, ColumnConfigs, WithID } from "./types"
+
+interface Props<Datum extends WithID> {
+  data: Datum[]
+  columnConfigs: ColumnConfigs<Datum>
+}
+
+export function Table<Datum extends WithID>(props: Props<Datum>) {
+  const { data: dataUnsorted, columnConfigs } = props
+  const [sortConfig, setSortConfig] = useState<ColumnConfig<Datum>>()
```
luiz-chagas: Name nitpick ⛏️ 

`sortConfig` threw me off a little bit, I initially thought there was an explicit place for a `SortConfig` and then realized it came from the `ColumnConfig` (as the type suggests).

`sort` is the feature that uses the type `ColumnConfig`, so we could infer a variety of names to represent this:
`sortColumnConfig`, `sortColumn`, `sortConfig`. This last one seems the least obvious to me for some reason, maybe it put me on a mindset of a configuration object, I don't know 🤷🏼

src/components/Table/types.ts
```diff
@@ -0,0 +1,13 @@
+import { ReactNode } from "react"
+
+export interface WithID {
+  id: string
+}
+
+export interface ColumnConfig<Datum extends WithID> {
+  label: string | null
+  getValue?: (datum: Datum) => ReactNode
```
luiz-chagas: I think there's an opportunity here to make `getValue` a bit more strict by forcing its return type to be `string | number | boolean` and the benefit is that these values are also sortable, which would let you get rid of that `isSortable` check elsewhere.

The drawback is that one wouldn't be able to shove any JSX into the table

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  const handleSort = (columnConfig: ColumnConfig<Datum>) => () => {
+    if (columnConfig.sortable && columnConfig.getValue) {
+      if (columnConfig === sortConfig || !sortConfig) {
+        setSortAsc((value) => !value)
+      }
+      setSortConfig(columnConfig)
+    }
+  }
```
luiz-chagas: Any thoughts on `if` nesting? It's pretty easy to follow in this case but I find that nesting becomes hard to track when functions are on the larger side.

However, there's usually a way to invert the conditions and remove nesting:
```tsx
if (!columnConfig.sortable || !columnConfig.getValue) return;

if (columnConfig === sortConfig || !sortConfig) {
    setSortAsc((value) => !value)
}

setSortConfig(columnConfig)
```

luiz-chagas (dismissed):
I think this is looking good, I left a bunch of ramblings but I wouldn't call any of them issues.

src/components/Table/types.ts
roblafeve: Yep, wondered about this—I think I tend to agree 👍

roblafeve: Yep, I think this is a useless abstraction 👌

src/components/Table/Table.tsx
roblafeve: In my initial thinking, I was trying to accomodate filler columns—but now that I've played with this a bit, I'm not seeing that as a likely scenario. Making these required will simplify the implementation and usage. Good catch! 👍

src/components/Table/Table.tsx
roblafeve: Ha ha, yeah—I hear you. I went back and forth on this one. I think the term "config" throws me off a bit. I wonder if it would be odd to simply change the type `ColumnConfig` to `Column`. What do you think?

I struggled with this since a column might be thought of as the one-dimensional Row > Column... but what I intend here is a Table > Column. But the word "config" thrown into the mix doesn't necessarily help anything.

src/components/Table/types.ts
roblafeve: Great catch—and it gets at something I've been thinking about. I think we do need a selector function (`getValue`) that gets the primitive value—so yeah, let's tighten this up. I think an optional second function could be provided that would then be passed that value to be interpolated into some JSX structure. We need `getValue` for sorting, but we can't sort JSX... so let's separate them.

src/components/Table/Table.tsx
roblafeve: That's smart 👌

src/components/Table/types.ts
thispastwinter: Yeah, I think @luiz-chagas makes a great point. Enforcing the return value is a nice way to avoid the type checking. Then you can just have a separate render callback that also returns the value to be rendered within jsx.

src/components/Table/Table.tsx
thispastwinter: @luiz-chagas I love this. I see this in your code often and really appreciate how clean this is. For some reason I always avoid empty returns, but it is such a clean way of escaping the function.

src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+  { id: "1", name: "Joe", age: 43 },
+  { id: "4", name: "Andrea", age: 34 },
+  { id: "3", name: "Rob", age: 37 },
+  { id: "2", name: "Tim", age: 40 },
+]
+
+export const Example = () => (
+  <Table
+    data={data}
+    columnConfigs={[
```
thispastwinter: What if we just called this columns? Maybe it's not as accurate of a name...but I don't like the name columnConfigs. Total nitpick ⛏️  and feel free to ignore my rude feedback.

roblafeve: I COMPLETELY AGREE! 😆

src/components/Table/types.ts
roblafeve: Thinking through this a bit more—it could be possible that you may want to configure a column that doesn't require any `Datum` information and only uses the `renderCell` (or whatever we call it) function to render something. In this case, if `getValue` was required, the user would still need to define it and return a `string | number | boolean` just to ignore it in `renderCell`.

I'm going to think through some options here.

thispastwinter: Maybe `sortable` can be used to discriminate whether `getValue` is required 😄

thispastwinter: Or, maybe rethink it entirely. Maybe instead of `getValue` it's `sortBy`

roblafeve: > Or, maybe rethink it entirely. Maybe instead of `getValue` it's `sortBy`

@thispastwinter I think you're onto something there. I'm not sure we can type what I've been envisioning. It would be much easier to just pass the entire item into both functions. Thanks for helping me think outside the box I was stuck in! 🙏

thispastwinter: I think the nice thing about sortBy is that it removes the need for a Boolean.

src/components/Table/README.md
```diff
@@ -0,0 +1,120 @@
+# `<Table />`
```
thispastwinter, line 1: I really like this documentation. Super love not using tables in markdown!

src/components/Table/Table.module.css
```diff
@@ -0,0 +1,25 @@
+.table {
```
thispastwinter: Should we avoid using table here since our component name is `<Table />` and our css is scoped?

src/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+    columns={[
+      {
+        id: "name",
+        label: "Name",
+        renderCell: ({ name }) => name,
+      },
+      {
+        id: "age",
+        label: "Age",
+        sortBy: ({ age }) => age,
```
thispastwinter, line 58: 🥳

src/components/Table/Table.module.css
roblafeve: 😉 ha ha, yeah, probably

thispastwinter:
At this point are we just waiting for the `Row selection` piece to be implemented?

roblafeve:
@thispastwinter I think I'll plan on coming back to this next Lab Friday, if you'll have me 🥲

PR #49, opened 2022-08-24
Feature/avatar list

### LINKED ISSUE #54: Avatar List

thispastwinter:
## Description

A wrapper for a group of `<Avatar />`'s. This component's default implementation assumes a minimal list to be passed in. If your use case requires hundreds if not thousands of user avatars, you may want to pass in your own hidden count element via `renderHiddenCount` for handling custom display logic. Please note that this component will throw an error if it receives any children that are not of type `<Avatar />`. Additionally, the size that is passed into the parent `<AvatarGroup />` will override any of its children.

### 🔗 Reference

- [LINK_NAME](LINK_URL)

### 📋 Acceptance Criteria

- [ ] It should only render `<Avatar />` children
- [ ] A size prop will render each `<Avatar />` in the given size that is supplied
- [ ] The size will override any of the sizes specified by its children
- [ ] It allows the rendering of a custom hidden element

### 🖼 Demo
Screen Shot 2022-08-29 at 10 04 07 AM

### 👩‍🔬 Test Instructions

1. TEST_STEP_HERE

## 🔎 Reviewer Checklist

- [x] Merge destination is correct

---

>
> #### 👨‍💻 Assignee
>
>
> #### 👩‍💻 Reviewer
>


roblafeve:
@thispastwinter I think my main question here is that of organization. I think `AvatarGroup` could be located within the existing `Avatar` module folder and exported from there. These components are highly-related and it would allow for more natural sharing of styling between `Avatar` and the "count" element in `AvatarGroup`. Thoughts?

src/components/Avatar/Avatar.tsx
```diff
@@ -1,38 +1,27 @@
...
-export function Avatar({ name, size = "sm", src }: AvatarProps) {
+export function Avatar({ name, ...avatarBaseProps }: AvatarProps) {
   const Child = () => {
-    if (src) {
+    if (avatarBaseProps.src) {
       return null
     } else {
       const initials = getInitials(name)
-      return <div className={styles.text}>{initials}</div>
+      return <span>{initials}</span>
```
roblafeve: Seems like this `<span>` doesn't serve any purpose and it renders fine without it.

src/components/Avatar/index.ts
```diff
@@ -1 +1,3 @@
-export { Avatar, type AvatarSize } from "./Avatar"
+export { Avatar, type AvatarProps } from "./Avatar"
```
roblafeve, line 1: 💭 We should get on the same page on whether or not we export component prop interfaces. I can understand how it could be valuable in some instances—extending a component—so I'm not sayinig we shouldn't.

src/components/Avatar/AvatarBase/AvatarBase.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export type AvatarSize = "sm" | "md" | "lg"
+
+export interface AvatarBaseProps {
+  size?: AvatarSize
+  children: ReactNode
+  ariaLabel?: string
+  src?: string
+  fontColor?: string
+  backgroundColor?: string
```
roblafeve: Seeing this fresh—I wonder if it would be more flexible to pass `background` instead of `background-color`. Maybe there's a world where a background gradient is desired?

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  hiddenCount,
+  renderHiddenCount,
+  size,
+  total,
+}: HiddenCountProps) {
+  if (!hiddenCount) return null
+
+  if (renderHiddenCount) {
+    return renderHiddenCount(hiddenCount)
+  } else {
```
roblafeve: We can nix the `else` in this case and reduce nesting

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return (
+    <div className={styles.container}>
+      {slicedChildren.map(({ props }, index) => (
+        <div
+          data-testid={`avatar_item_${index}`}
+          key={index}
+          className={styles.item}
+          style={{ zIndex: 1000 - index }}
```
roblafeve: An arbitrarily high z-index like this is prone to conflict with other layered elements (e.g. an overlay or modal). I think we could use the `index` as-is for the z-index. Then in your styles for `.container` you can set position to `relative` to ensure that context for this ordering is done in relation to this component. Then in the `Count` component, you could do `total - 1` to ensure that the `z-index`es exactly match the intended sequence.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  children,
+  renderHiddenCount,
+}: AvatarGroupProps) {
+  const childrenArray = Array.isArray(children) ? children : [children]
+  const slicedChildren = childrenArray.slice(0, max)
+  const hiddenCount = childrenArray.length - slicedChildren.length
+
+  if (childrenArray.some((child) => child.type.name !== "Avatar")) {
+    throw new Error("<AvatarGroup /> only accepts <Avatar /> children")
+  }
```
roblafeve, line 29: Nice 👌

src/components/Avatar/AvatarGroup/AvatarGroup.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <Avatar name="Jane Doe" />
+          <Avatar name="Foo Bar" />
+        </AvatarGroup>,
+      )
+
+    // Assert
+    expect(Rendered).toThrowError(
+      "<AvatarGroup /> only accepts <Avatar /> children",
+    )
+    jest.restoreAllMocks()
```
roblafeve, line 45: I'm not doing this for Tabs 😬—good job!

src/components/Avatar/AvatarGroup/AvatarGroup.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    })
+  })
+
+  it("does not render hidden count element if number of children equals max", () => {
+    // Arrange
+    const testId = "hidden_count"
+
+    // Act
+    render(
+      <AvatarGroup>
```
roblafeve: 💭 Maybe we should explicitly pass `max` here since this test now depends on the default not changing?

roblafeve (commented):
This is looking good 👏

src/components/Avatar/AvatarBase/AvatarBase.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <AvatarBase background={backgroundColor} fontColor={fontColor}>
+        {name}
+      </AvatarBase>,
+    )
+
+    const received = screen.getByText(name)
+
+    // Assert
+    expect(received).toHaveStyle(
+      `color: ${fontColor}; background-color: ${backgroundColor};`,
```
roblafeve, line 47: Interesting—you're testing for `background-color` but I notice that you're passing `background` in the implementation—but it is still passing. I wonder why it still passes?

src/components/Avatar/AvatarBase/AvatarBase.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      // Assert
+      expect(received).toHaveAttribute("data-size", size)
+    })
+  })
+
+  it("allows for override of background and font color", () => {
+    // Arrange
+    const name = "Crema Components"
+    const fontColor = "rgb(255, 0, 0)"
+    const backgroundColor = "rgb(0, 0, 0)"
```
roblafeve, line 30: ⛏️ I could see renaming this value to `background` to align with the prop...

src/components/Avatar/AvatarBase/AvatarBase.module.css
```diff
@@ -2,15 +2,12 @@
...
-  background: #999999;
-  color: #333333;
+  --ratio: 0.4;
   display: flex;
   align-items: center;
   justify-content: center;
   border-radius: 100%;
   flex-wrap: nowrap;
-  background-size: cover;
 }
```
roblafeve, line 11: Is the removal of `background-size: cover;` intentional? It seems like this would necessitate the user to pass exactly square images.

Maybe I'm missing something...

src/components/Avatar/AvatarBase/AvatarBase.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReactNode } from "react"
+import styles from "./AvatarBase.module.css"
+
+export type AvatarSize = "sm" | "md" | "lg"
+
+export interface AvatarBaseProps {
+  size?: AvatarSize
+  children: ReactNode
+  ariaLabel?: string
```
roblafeve: Maybe `ariaLabel` should be required?

src/components/Avatar/AvatarBase/AvatarBase.tsx
```diff
@@ -0,0 +1,37 @@
...
+  fontColor: color = "#444444",
+  background = "#b3b3b3",
+}: AvatarBaseProps) {
+  return (
+    <div
+      aria-label={ariaLabel}
+      style={{
+        color,
+        background: src ? `url(${src})` : background,
+        backgroundSize: "cover",
```
roblafeve, line 29: Ah, here's where `background-size` moved to. Since this property doesn't need to be dynamic, I could see moving it back to the stylesheet.

src/components/Avatar/AvatarGroup/AvatarGroup.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+]
+
+export default {
+  title: "Components/AvatarGroup",
+  argTypes: {
+    size: {
+      control: "select",
+      options: ["sm", "md", "lg"] as AvatarProps["size"][],
+    },
+  },
```
roblafeve, line 28: 👍 I need to use this more...

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,82 @@
+import { ComponentElement, ComponentState } from "react"
+import { Avatar, AvatarProps } from ".."
+import { AvatarBase } from "../AvatarBase"
+import styles from "./AvatarGroup.module.css"
+
+type AvatarGroupChildren =
+  | ComponentElement<AvatarProps, ComponentState>
+  | ComponentElement<AvatarProps, ComponentState>[]
```
roblafeve, line 8: Does this provide value over:
```tsx
type AvatarGroupChildren = 
  | ReactElement<AvatarProps>
  | ReactElement<AvatarProps>[]
```
Just curious.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export function AvatarGroup({
+  size = "sm",
+  max = 3,
+  children,
+  renderHiddenCount,
+}: AvatarGroupProps) {
+  const childrenArray = Array.isArray(children) ? children : [children]
+  const slicedChildren = childrenArray.slice(0, max)
```
roblafeve: I wonder if `visibleChildren` or something more descriptive would be more helpful in understanding what we intend to achieve with slice.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  return (
+    <div className={styles.container}>
+      {slicedChildren.map(({ props }, index) => (
+        <div
+          data-testid={`avatar_item_${index}`}
+          key={index}
+          className={styles.item}
+          style={{ zIndex: index }}
```
roblafeve: Seeing this fresh—currently we don't need to specify `z-index` as the sequence will naturally stack as intended based on document order. Missed that the first time around. 😆

src/components/Avatar/README.md
```diff
@@ -18,8 +18,10 @@ A component for rendering a circular container used to display a user's initials
 
 ## API
```
roblafeve, line 20: ...Whole file...

I wonder if this README should serve both `Avatar` and `AvatarGroup`. Maybe not... but maybe we can at least link to `AvatarGroup`s README from here.

roblafeve (dismissed):
This looks solid! I don't think any of my feedback is a show-stopper.

src/components/Avatar/AvatarBase/AvatarBase.module.css
thispastwinter: It's still present in component via the style prop. It seemed to be the only way I could get background-size to work with the modified background implementation.

src/components/Avatar/AvatarBase/AvatarBase.tsx
thispastwinter: Ah yes. In this instance it should be since it's omitted in the main component.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+  children,
+  renderHiddenCount,
+}: AvatarGroupProps) {
+  const childrenArray = Array.isArray(children) ? children : [children]
+  const slicedChildren = childrenArray.slice(0, max)
+  const hiddenCount = childrenArray.length - slicedChildren.length
+
+  if (childrenArray.some((child) => child.type.name !== "Avatar")) {
+    throw new Error("<AvatarGroup /> only accepts <Avatar /> children")
+  }
```
luiz-chagas, line 29: Is this a given with the typing of `AvatarGroupChildren`?

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface HiddenCountProps {
+  hiddenCount: number
+  size: AvatarGroupProps["size"]
+  renderHiddenCount: AvatarGroupProps["renderHiddenCount"]
+  total: number
+}
+
+function HiddenCount({
+  hiddenCount,
+  renderHiddenCount,
```
luiz-chagas: Naming: hiddenCount is both the component and a prop. There's also a repetition of `hiddenCount`. If this is the `HiddenCount` component, do props also need to inform that?

src/components/Avatar/Avatar.tsx
```diff
@@ -1,38 +1,28 @@
...
   const Child = () => {
-    if (src) {
+    if (avatarBaseProps.src) {
       return null
     } else {
       const initials = getInitials(name)
-      return <div className={styles.text}>{initials}</div>
+      return <>{initials}</>
     }
   }
```
luiz-chagas: Naming: Child seems too generic, everything is a child in react. Are there simpler alternatives to get this implemented?
This might not be simpler, but one idea would be to replace Child with an expression:
```tsx
{ hasImage ? null : getInitials(name) }
```
Or if you don't want to use ternaries
```tsx
{ doesNotHaveImage && getInitials(name) }
```

luiz-chagas (commented):
Combining `AvatarGroup`, `AvatarBase` and `Avatar` into the same module is a great solution for all the interdependencies without violating any privacy rules 👍🏼

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
thispastwinter: Ah yeah, good catch.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
thispastwinter: The reason I wasn't using ReactElement is that it doesn't include the name property on `type` which you can see I'm using below where the error is thrown.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
thispastwinter: So yes it is, however you don't get the type inference alone and someone could abuse the api very easily and just pass in whatever they wanted.

src/components/Avatar/AvatarBase/AvatarBase.tsx
roblafeve: We talked about the complexity here given the specificity of `background` inline and decided to stick with `backgroundSize` being inline.

src/components/Avatar/AvatarBase/AvatarBase.module.css
roblafeve: Sticking with current implementation.

src/components/Avatar/AvatarGroup/AvatarGroup.tsx
roblafeve: That's smart 👍

src/components/Avatar/AvatarBase/AvatarBase.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    expect(received).toHaveStyle(
+      `color: ${fontColor}; background-color: ${backgroundColor};`,
+    )
+  })
+
+  it("renders an image when a src is passed", () => {
+    // Arrange
+    const fileName = "justink.jpeg"
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const image = require(`../../../assets/${fileName}`)
```
belv2c, line 55: Instead of using require I think we could do `import justink from "../../../assets/justink.jpeg"`

src/components/Avatar/AvatarGroup/AvatarGroup.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * See Storybook Docs: Writing Stories
+ * https://storybook.js.org/docs/basics/writing-stories/
+ */
+
+const avatars: AvatarProps[] = [
+  { name: "Crema Components" },
+  { name: "Hello World", size: "md" },
+  {
+    name: "Foo Bar",
+    src: require("../../../assets/justink.jpeg"),
```
belv2c: Same import vs require comment from above

roblafeve (dismissed):
Good work!

belv2c (approved):
ᕕ( ᐛ )ᕗ

